PR #15, opened 2019-03-27
Update for <= k8s 1.12 kubernetes versions

LKaemmerling:
The cloud controller is now useable with <= k8s 1.12. We should release this as 1.3.0.

This was tested on K8s 1.12.


hcloud/instances.go
```diff
@@ -88,15 +88,15 @@ func (i *instances) InstanceTypeByProviderID(providerID string) (string, error)
...
 	return errors.New("not implemented")
 }
 
-func (i *instances) CurrentNodeName(hostname string) (types.NodeName, error) {
+func (i *instances) CurrentNodeName(ctx context.Context, hostname string) (types.NodeName, error) {
 	return types.NodeName(hostname), nil
 }
 
-func (i instances) InstanceExistsByProviderID(providerID string) (exists bool, err error) {
+func (i instances) InstanceExistsByProviderID(ctx context.Context, providerID string) (exists bool, err error) {
```
thcyron: Now that you have a context, shouldn’t you pass it around when calling the hcloud client?

```go
server, _, err = i.client.Server.GetByID(context.Background(), id)
```

(A few lines below.)

hcloud/util.go
```diff
@@ -57,7 +57,7 @@ func providerIDToServerID(providerID string) (id int, err error) {
 		return
 	}
 
-	idString := strings.Replace(providerID, providerPrefix, "", 1)
+	idString := strings.ReplaceAll(providerID, providerPrefix, "")
```
thcyron: Just to make sure: these two lines don’t do the same (first line replaces first occurrence, second line replaces all occurrences).

hcloud/zones.go
```diff
@@ -31,7 +33,8 @@ func newZones(client *hcloud.Client, nodeName string) cloudprovider.Zones {
 	return zones{client, nodeName}
 }
 
-func (z zones) GetZone() (zone cloudprovider.Zone, err error) {
+func (z zones) GetZone(ctx context.Context) (zone cloudprovider.Zone, err error) {
+	log.Print("GetZone")
```
thcyron: Do you want to keep those logs?

hcloud/util.go
LKaemmerling: This is intended.

hcloud/instances.go
```diff
@@ -34,84 +34,100 @@ func newInstances(client *hcloud.Client) *instances {
...
 
-func (i *instances) NodeAddressesByProviderID(providerID string) ([]v1.NodeAddress, error) {
+func (i *instances) NodeAddressesByProviderID(ctx context.Context, providerID string) ([]v1.NodeAddress, error) {
 	id, err := providerIDToServerID(providerID)
 	if err != nil {
 		return nil, err
 	}
 
-	server, err := getServerByID(i.client, id)
+	server, err := getServerByID(i.client, id, ctx)
```
thcyron: FYI: It’s a generally-accepted convention to pass `ctx` always as the first argument.

PR #17, opened 2019-05-11
Fix node deletion for removed instances

Nighoo:
Deleting a running hetzner node will cause a nil pointer dereference when InstanceShutdownByProviderID is called for that node. This is caused by the fact that GetByID will return nil and not an error for the requested server.

### DIFF
--- a/hcloud/instances.go
+++ b/hcloud/instances.go
@@ -18,7 +18,7 @@ package hcloud
 
 import (
 	"context"
-	"errors"
+	"k8s.io/kubernetes/pkg/cloudprovider"
 	"strconv"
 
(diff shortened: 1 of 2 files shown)


Nighoo:
I adjusted the commit to fix the return value for removed instances. InstanceShutdownByProviderID must not return true for deleted instances. So now the kubernetes nodes are correctly removed when the cloud instances are deleted.
In addition i fixed the Instance ID to return only the server id, since the provider name is added by GetInstanceProviderID. (see kubernetes/cloud-provider cloud.go)

LKaemmerling:
LGTM thank you!

PR #21, opened 2019-07-18
Set node InternalIP from attached networks

### LINKED ISSUE #19: Add Networks support

hwuethrich:
Fixes #19 by setting the node `InternalIP` to the attached network IP address.

I could not find any information about wether multiple addresses of type `InternalIP` are permitted or not. The controller is currently adding all IPs if the node is attached to multiple networks.

(Also, I also optimised the docker container build time. Hope this is ok.)


LKaemmerling:
Hello,

first of all thank you! We have this already in a sperate branch - which was just released as a beta a few moments ago (https://github.com/hetznercloud/hcloud-cloud-controller-manager/issues/19#issuecomment-512850269). 

Setting only the `InternalIP` is not enough :). If you want you can test our implementation.

PR #26, opened 2019-10-11
Correct wrong README link in documentation

oveits:
I have updated a link, so yoiu do not get a 404 Not Found anymore...


LKaemmerling:
Thank you!

PR #72, opened 2020-08-06
Implement change load balancer type

fhofherr:


fhofherr:
> Keep in mind to add a changelog entry :)

Done

PR #83, opened 2020-08-29
Fix secret creation in deploy with networks guide.

yeldiRium:
The previous two steps could not work since they tried to create the same secret twice. The secret must be created in one step.


LKaemmerling:
Thank you!

Issue #99, opened 2020-10-23
Incorrect instance ID returned 


Fgruntjes:
When a node has been added to the cluster without setting the value in `node.Spec.ProviderID`, then the `InstanceID` will be called and returns only the `<hetzner server id>` without the prefix. The other methods are expecting the ID in the form of `hcloud://<hetzner server id>` and will throw an error.

```
E1023 14:35:34.330136   95896 node_controller.go:245] Error getting node addresses for node "cx41-fsn1-5486140987150761883": error fetching node by provider ID: hcloud/instances.NodeAddressesByProviderID: hcloud/providerIDToServerID: missing prefix hcloud://: , and error by node name: hcloud/instances.NodeAddresses: instance not found
```

LKaemmerling:
The cluster needs to be created with the config `Environment="KUBELET_EXTRA_ARGS=--cloud-provider=external"`  (https://github.com/hetznercloud/hcloud-cloud-controller-manager/issues/80). K8s will set the "prefix" automatically.

Fgruntjes:
Sorry for me being blind, thank you!

PR #100, opened 2020-10-23
Updated the return value of the InstanceID method to include hcloud prefix

Fgruntjes:
Should solve the issue described in #99


Fgruntjes:
As stated in #99 this was my mistake

PR #106, opened 2020-10-28
Add option to keep server after e2e test failure

fhofherr:
This allows manual analysis of the environment after a test failure.

### DIFF
--- a/scripts/e2etest-local.sh
+++ b/scripts/e2etest-local.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+
+set -e
+
+function test_k8s_version() {
+    if [[ -z "$1" ]]; then
+        echo "Usage: $0 <k8s-version>"
+        return 1
+    fi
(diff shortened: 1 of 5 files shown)


scripts/e2etest-local.sh
```diff
@@ -0,0 +1,36 @@
...
+        return 2
+    fi
+}
+
+if [[ -z "$HCLOUD_TOKEN" ]]; then
+    echo "HCLOUD_TOKEN not set! Aborting tests."
+    exit 1
+fi
+
+K8S_VERSIONS=("1.17.12" "1.18.9" "1.19.3")
```
LKaemmerling, line 33: I think we should have "one" single source of truth for our supported k8s versions. Otherwise, we end up changing the values on 3-4 different places (`gitlab-ci`, `github-actions`, `README.md`, this file)

LKaemmerling (approved):

PR #114, opened 2020-11-10
Add load balancer defaults via env

0SkillAllLuck:
Add env for: 
- HCLOUD_LOAD_BALANCERS_DEFAULT_LOCATION
- HCLOUD_LOAD_BALANCERS_DEFAULT_NETWORK_ZONE
- HCLOUD_LOAD_BALANCERS_DEFAULT_USE_PRIVATE_IP
- HCLOUD_LOAD_BALANCERS_DEFAULT_DISABLE_PRIVATE_INGRESS

HCLOUD_LOAD_BALANCERS_DEFAULT_LOCATION and HCLOUD_LOAD_BALANCERS_DEFAULT_NETWORK_ZONE help with the error of "neither load-balancer.hetzner.cloud/location or load-balancer.hetzner.cloud/network-zone are set".
This removes the need to set these annotations over and over again if the cluster is only in a single location.

### DIFF
--- a/internal/hcops/load_balancer.go
+++ b/internal/hcops/load_balancer.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 	"strconv"
 	"strings"
 	"sync"
@@ -13,6 +14,12 @@ import (
 	"github.com/hetznercloud/hcloud-go/hcloud"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/klog/v2"
+)
+
+const (
+	hcloudLoadBalancersDefaultLocationENVVar     = "HCLOUD_LOAD_BALANCERS_DEFAULT_LOCATION"
+	hcloudLoadBalancersDefaultNetworkZoneENVVar  = "HCLOUD_LOAD_BALANCERS_DEFAULT_NETWORK_ZONE"
+	hcloudLoadBalancersDefaultUsePrivateIPENVVar = "HCLOUD_LOAD_BALANCERS_DEFAULT_USE_PRIVATE_IP"
 )
 
 // HCloudLoadBalancerClient defines the hcloud-go functions required by the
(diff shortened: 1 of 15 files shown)


internal/hcops/load_balancer.go
```diff
@@ -333,8 +344,21 @@ func (l *LoadBalancerOps) ReconcileHCLBTargets(
 	)
 
 	usePrivateIP, err := annotation.LBUsePrivateIP.BoolFromService(svc)
-	if err != nil && !errors.Is(err, annotation.ErrNotSet) {
-		return changed, fmt.Errorf("%s: %w", op, err)
+	if err != nil {
+		if errors.Is(err, annotation.ErrNotSet) {
+			klog.Info("Not set")
```
LKaemmerling, line 349: Could you please remove the debug line?

internal/hcops/load_balancer.go
```diff
@@ -333,8 +344,21 @@ func (l *LoadBalancerOps) ReconcileHCLBTargets(
...
+			if usePrivateIPDefault, ok := os.LookupEnv(hcloudLoadBalancersDefaultUsePrivateIPENVVar); ok {
+				if usePrivateIP, err = strconv.ParseBool(usePrivateIPDefault); err != nil {
+					klog.Info("to bool failed")
```
LKaemmerling, line 352: This too

internal/hcops/load_balancer.go
```diff
@@ -333,8 +344,21 @@ func (l *LoadBalancerOps) ReconcileHCLBTargets(
...
+					return changed, fmt.Errorf("%s: %w", op, err)
+				}
+				klog.Info("ok")
```
LKaemmerling, line 355: Too

internal/hcops/load_balancer.go
```diff
@@ -333,8 +344,21 @@ func (l *LoadBalancerOps) ReconcileHCLBTargets(
...
+			} else {
+				klog.Info("no default")
```
LKaemmerling, line 357: Too :)

LKaemmerling (changes requested):
Hey @0SkillAllLuck, thank you for your contribution! I have a few smaller things. Could you also please add test cases to `internal/hcops/load_balancer_test.go`?

LKaemmerling:
@0SkillAllLuck could you please rebase against master? Then the tests should be executed.

0SkillAllLuck:
Added HCLOUD_LOAD_BALANCERS_DEFAULT_DISABLE_PRIVATE_INGRESS to potentially help with #118. When only the public ingress is needed.

0SkillAllLuck:
@LKaemmerling rebased against master.

LKaemmerling:
@0SkillAllLuck we were thinking about this a lot. 
It looks like there are a few changes in there that shouldn't be there :) Also, we think that the ENV names are really too long and not documented at all, so we would like to come up with our own solution within the next couple of weeks. Therefore, thank you for your work on this but unfortunately, we decided to reject this merge request.

PR #144, opened 2021-01-10
Add deployment yaml to use only internal ip assignment of nodes

NotANormalNerd:
This deployment yaml allows to to use the cloud controller to assign the internal IPs for the nodes without using the whole pod cidrs network features.

In combination with rancher this finally allows for a working private network usage without any hacks.

### DIFF
--- a/deploy/ccm-internal-ips.yaml
+++ b/deploy/ccm-internal-ips.yaml
@@ -1,0 +1,82 @@
+# NOTE: this release was tested against kubernetes v1.18.x
+
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: cloud-controller-manager
+  namespace: kube-system
+---
+kind: ClusterRoleBinding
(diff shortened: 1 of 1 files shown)


LKaemmerling:
Hey @NotANormalNerd,

with this change, you basically lose all the benefits from the Networks. The Cloud Controller Manager won't configure the network correctly anymore and it basically works like deployed without Networks. You can always use this deployment for yourself, but we won't recommend it, therefore I'm sorry to tell you that we won't accept this MR. 

Thank you!

NotANormalNerd:
Guten Morgen @LKaemmerling,

Thanks for the response. In my case the node is deployed with Rancher and the https://github.com/JonasProgrammer/docker-machine-driver-hetzner, which connects the node with the network on deployment.

I think in that usecase is not well documented but I will create a seperate repository/documentation for that usecase.

Thanks for the work :-D

PR #155, opened 2021-02-16
Api deprecations in deployment yamls

ekeih:
rbac.authorization.k8s.io/v1beta1 has been deprecated since Kubernetes 1.17 and will be removed with 1.22. This change does not change the compatibility of the cloud controller manager as it only touches the current example deployment yaml files.

While I was at it I noticed the broken links in the readme, so I took the liberty to fix those in the same PR (but separate commit).


LKaemmerling (approved):

LKaemmerling:
Thank you! Good job!

PR #187, opened 2021-04-24
Fix version tag in deploy folder

aspettl:
Issue: The latest release 1.9.0 still references 1.8.1 in the deployment yaml files.

This PR would fix it at least on the master branch (so that the instructions in the readme file pull in the latest version, which is the expectation). But probably 1.9.1 should be released instead so that the git tag can also contain this change (then with v1.9.1, obviously). So feel free to reject this PR if you want to go for the 1.9.1 release directly.

### DIFF
--- a/deploy/ccm-networks.yaml
+++ b/deploy/ccm-networks.yaml
@@ -54,7 +54,7 @@ spec:
           effect: "NoSchedule"
       hostNetwork: true
       containers:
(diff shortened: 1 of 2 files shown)


fhofherr (approved):

fhofherr:
Hi @aspettl,

thank you for your pull request. We will release a version 1.9.1 soon. However, I will merge your PR so that anyone stumbling over this in the mean time gets the updated config.

Sorry for the inconvenience.

PR #194, opened 2021-05-05
Add option to disable IPv6 for load balancers

### LINKED ISSUE #191: Disabling IPv6 for loadbalancers
Hello,
Is there an option "annotation/flag" to disable IPv6 Support ? 

I have upgraded CCM to v1.9.1 but the external-dns erroring to : the target A is not a valid record. As there is an open issue with external-dns regarding IPv6 support  kubernetes-sigs/external-dns/issues/2044 

I understand the feature was re-enabled #137 , but the ingress-nginx-controller is always publishing both IPv4 and IPv6 even with ConfigMap -> disable-ipv6-dns: "true" 

The only workaround I found is to go back to v1.8.1 where IPv6 is completely not supported.

fhofherr:
Enabling IPv6 by default in #137 broke clusters using external-dns. This
commit adds an option to disable IPv6 by setting the

    load-balancer.hetzner.cloud/ipv6-disabled

annotation to true.

Closes #191

### DIFF
--- a/hcloud/load_balancers.go
+++ b/hcloud/load_balancers.go
@@ -59,14 +59,23 @@ func (l *loadBalancers) GetLoadBalancer(
 		}, true, nil
 	}
 
-	return &v1.LoadBalancerStatus{Ingress: []v1.LoadBalancerIngress{
+	ingresses := []v1.LoadBalancerIngress{
 		{
 			IP: lb.PublicNet.IPv4.IP.String(),
 		},
-		{
+	}
+
+	disableIPV6, err := annotation.LBIPv6Disabled.BoolFromService(service)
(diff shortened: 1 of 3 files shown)


LKaemmerling:
I would invert the annotation name, because otherwise you have Double „no“. (EnableIPv6 that defaults to true is better than having DisableIPv6 that defaults to false and when it is false it disables IPv6)

fhofherr:
> I would invert the annotation name, because otherwise you have Double „no“. (EnableIPv6 that defaults to true is better than having DisableIPv6 that defaults to false and when it is false it disables IPv6)

Only the commit message was wrong. The annotation has to be set to true of course to disable ipv6.

LKaemmerling (approved):

PR #208, opened 2021-06-04
Add Master is attached to configured Network check on controller boot.

LKaemmerling:
We had several reports/issues from customers who configured there CCM wrong. They copied there configuration and changed nothing (It uses the same API token and the configured network as the other cluster). However, the servers of the new cluster were in a different project and/or attached to a different network. Because of this, two CCM instances started to fight against each other, because the CCM always tries to reach the wanted state in the network, so it detaches everything it does not know about and then attaches his "known" routes.

With this MR we now check if the master is attached to the configured network by using a "not user controllable" source of truth - our metadata service, that is only reachable through the node public interface. This check will be performed on boot and if the node is not attached to the correct network, it will fail booting up with a message that gives this hint. I also added a new ENV to disable this check for cluster specific reasons (sample: Rancher clusters where the controlplan might be out of the cluster itself). We decided to check this "per default", so there might be issues for customers who have currently a somehow broken cluster but didn't noticed it.

Signed-off-by: Lukas Kämmerling <lukas.kaemmerling@hetzner-cloud.de>

### DIFF
--- a/e2etests/testing.go
+++ b/e2etests/testing.go
@@ -120,7 +120,7 @@ func (tc *TestCluster) initialize() error {
 		err := runCmd(
 			"go",
 			[]string{"build", "-o", "../hcloud-cloud-controller-manager", "../."},
-			[]string{"CGO_ENABLED=0"},
+			[]string{"CGO_ENABLED=0", "GOOS=linux", "GOARCH=amd64"},
 		)
 		if err != nil {
 			return fmt.Errorf("%s: %v", op, err)
--- a/hcloud/cloud.go
+++ b/hcloud/cloud.go
@@ -21,8 +21,11 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
+	"net/http"
 	"os"
 	"strconv"
+	"strings"
 
 	"github.com/hetznercloud/hcloud-cloud-controller-manager/internal/hcops"
 	"github.com/hetznercloud/hcloud-go/hcloud"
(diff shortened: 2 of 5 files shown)


e2etests/testing.go
```diff
@@ -120,7 +120,7 @@ func (tc *TestCluster) initialize() error {
 		err := runCmd(
 			"go",
 			[]string{"build", "-o", "../hcloud-cloud-controller-manager", "../."},
-			[]string{"CGO_ENABLED=0"},
+			[]string{"CGO_ENABLED=0", "GOOS=linux", "GOARCH=amd64"},
```
mfrister, line 123: Why do we need this change?

hcloud/cloud.go
```diff
@@ -91,6 +96,15 @@ func newCloud(config io.Reader) (cloudprovider.Interface, error) {
 			return nil, fmt.Errorf("%s: Network %s not found", op, v)
 		}
 		networkID = n.ID
+		if networkDisableAttachedCheck := os.Getenv(hcloudNetworkDisableAttachedCheckENVVar); networkDisableAttachedCheck != "true" {
```
mfrister, line 99: We otherwise use `getEnvBool`, so we can give the user feedback on ambiguous configuration. I guess we should use it here as well.

hcloud/cloud.go
```diff
@@ -201,6 +215,23 @@ func loadBalancerDefaultsFromEnv() (hcops.LoadBalancerDefaults, bool, error) {
...
+	resp, err := http.Get("http://169.254.169.254/hetzner/v1/metadata/private-networks")
+	if err != nil {
+		return false, fmt.Errorf("%s: %s", op, err)
+	}
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return false, fmt.Errorf("%s: %s", op, err)
+	}
+	return strings.Contains(string(body), fmt.Sprintf("network_id: %d", networkId)), nil
```
mfrister, line 232:
```suggestion
	return strings.Contains(string(body), fmt.Sprintf("network_id: %d\n", networkId)), nil
```

I suggest adding at least the newline to the search string, so we can't accidentally match a network with a longer ID and the same prefix. Additionally, you could check for whitespace before `network_id`, so we don't accidentally match another field, in case we ever add one (e.g. `public_network_id`).

If you apply this, you'll want to check whether we actually have single newlines (`\n`) in the metadata.

Or, you could use the `gopkg.in/yaml.v2` package that's already in the CCM's dependencies to do this properly 😄

mfrister (commented):
Looks good, I've added a few suggestions for improvements.

I think we should document the new check and the env var somehwere.

e2etests/testing.go
LKaemmerling: @fhofherr added this a few weeks ago, until then I didn't run our E2E tests in the CCM, because otherwise I would noticed earlier that it builds a macos binary and tries to run it on a ubuntu server :) For @fhofherr it worked (out of the box) because he uses Linux amd64 as his development machine.

fhofherr (approved):

PR #219, opened 2021-07-09
add explanation for private node IPs

tobikris:
The documentation lacks with regards to the deployment without Networks support. If you want to manage networks yourself but still make use of the private networks in hcloud there is a special case.

It is necessary to let the CCM know about the network so it can add the node IPs properly.


LKaemmerling (approved):

LKaemmerling:
Hey @tobikris,

thank you!

PR #236, opened 2021-08-24
Adds opportunity to use ipv6 more flexible

### LINKED ISSUE #233: Global default for IPv6 on LBs
Hello,

as mentioned in the pkg docs, the annotation `load-balancer.hetzner.cloud/ipv6-disabled` can be used to remove the IPv6 address from a LoadBalancer service and as hinted there, it's pretty much required when running external-dns.
Would be nice to have an ENV var to disable the v6 address by default, similar to how `HCLOUD_LOAD_BALANCERS_DISABLE_PRIVATE_INGRESS`, gets rid of the private IP.

Thanks,
Malte

fynluk:
I implemented a feature to fix #233.

Now you can set a default with an environment variable named ```HCLOUD_LOAD_BALANCERS_DISABLE_IPV6```. 
But you still have the possibility to enable or disable ipv6 on individual services with following annotations:
```load-balancer.hetzner.cloud/ipv6-disabled```(noting new)
```load-balancer.hetzner.cloud/ipv6-enabled``` (If you would like to enable ipv6 on a specific service after you disabled ipv6 globally)

Signed-off-by: Fynn Späker <spaeker@23technologies.cloud>

### DIFF
--- a/hcloud/cloud.go
+++ b/hcloud/cloud.go
@@ -40,6 +40,7 @@ const (
 	hcloudLoadBalancersNetworkZone           = "HCLOUD_LOAD_BALANCERS_NETWORK_ZONE"
 	hcloudLoadBalancersDisablePrivateIngress = "HCLOUD_LOAD_BALANCERS_DISABLE_PRIVATE_INGRESS"
 	hcloudLoadBalancersUsePrivateIP          = "HCLOUD_LOAD_BALANCERS_USE_PRIVATE_IP"
+	hcloudLoadBalancersDisableIPv6		 = "HCLOUD_LOAD_BALANCERS_DISABLE_IPV6"
 	nodeNameENVVar                           = "NODE_NAME"
 	providerName                             = "hcloud"
 	providerVersion                          = "v1.9.1"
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by mstarostik** [comment]
Hey @fynluk,

thanks for taking this. From a quick test I can't confirm it's working though.
I had to
```
diff --git a/hcloud/load_balancers.go b/hcloud/load_balancers.go
index cef33ae4..f44a0e09 100644
--- a/hcloud/load_balancers.go
+++ b/hcloud/load_balancers.go
@@ -196,7 +196,7 @@ func (l *loadBalancers) EnsureLoadBalancer(
                }
                enableipv6Annotation, err := annotation.LBIPv6Enabled.BoolFromService(svc)
                if err != nil && !errors.Is(err, annotation.ErrNotSet) {
-                       return nil, false, fmt.Errorf("%s",err)
+                       return nil, fmt.Errorf("%s",err)
                }
                if !disableIPV6 || disableipv6Annotation || enableipv6Annotation {
                        ingress = append(ingress, v1.LoadBalancerIngress{IP: lb.PublicNet.IPv6.IP.String()})

```
to get it to build, but even then I always get the IPv6, regardless of what I set the environment variable or the annotations to.
Trying to see what's going on now.
Also, I'm not sure `load-balancer.hetzner.cloud/ipv6-disabled` and `load-balancer.hetzner.cloud/ipv6-enabled` are both needed as `load-balancer.hetzner.cloud/ipv6-disabled=false` whould do the trick.

LKaemmerling:
Hey @fynluk,

thank you for your request. We decided to use @mstarostik`s MR with this functionality. Thank you!

PR #237, opened 2021-08-24
Bring IPv6 flag in line with private ingress flag

mstarostik:
To implement my own feature request #233 and inspired by @fynluk's take in #236

Unfortunately #236 doesn't work for me, so here's my version.

This models the HCLOUD_LOAD_BALANCERS_DISABLE_IPV6 feature more closely
after how HCLOUD_LOAD_BALANCERS_DISABLE_PRIVATE_INGRESS is implemented.

@LKaemmerling @fhofherr I noticed EnsureLoadBalancer() returns the private ingress if enabled while GetLoadBalancer() does not. Is this intentional?


LKaemmerling (approved):

LKaemmerling:
Hey @mstarostik,

good job! I like the idea. I tested it locally and it works well. I also fixed the style tests.  

For EnsureLoadBalancer:
 I will have a look, on the first view it looks like this is wrong behavior, but as k8s always calls the GetLoadBalancer (even after creation) it shouldn't affect the cluster.

PR #246, opened 2021-09-19
Added annotation to select nodes by node label selector

steigr:
The idea and k8s related functions are from kubernetes-replicator:
https://github.com/mittwald/kubernetes-replicator/blob/master/replicate/common/strings.go#L34-L58

Then add `"load-balancer.hetzner.cloud/target-nodes-matching"` annotation to the service ( e.g. "node-role.kubernetes.io/worker" ). The list of nodes will then be filtered.

### DIFF
--- a/hcloud/load_balancers.go
+++ b/hcloud/load_balancers.go
@@ -4,9 +4,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"reflect"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
-	"reflect"
 
 	"github.com/hetznercloud/hcloud-cloud-controller-manager/internal/annotation"
 	"github.com/hetznercloud/hcloud-cloud-controller-manager/internal/hcops"
@@ -116,7 +117,7 @@ func MustGetObject(obj interface{}) metav1.Object {
 	panic(fmt.Errorf("Unknown type: %v", reflect.TypeOf(obj)))
 }
 
-func (l *loadBalancers) nodesForService(svc *v1.Service, nodes []*v1.Node) ([]*v1.Node) {
+func (l *loadBalancers) nodesForService(svc *v1.Service, nodes []*v1.Node) []*v1.Node {
 	objectMeta := MustGetObject(svc)
 	svcAnnotations := objectMeta.GetAnnotations()
 
@@ -129,7 +130,7 @@ func (l *loadBalancers) nodesForService(svc *v1.Service, nodes []*v1.Node) ([]*v
 		filteredNodes := []*v1.Node{}
 		for _, n := range nodes {
 			if nodeSelector.Matches(labels.Set(n.Labels)) {
-				filteredNodes = append(filteredNodes,n)
+				filteredNodes = append(filteredNodes, n)
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


alhafoudh:
Currently I have a problem that controller manager creates load balancer with targets of nodes which do not have nginx deployed. My nginx-ingress has nodeSelector defined to only include worker nodes. Does this solve the problem?

steigr:
Hi,

The CCM just assigns all nodes to the LB. If ingress-nginx does not run with HostPort then all nodes open the NodePort, routing to one of the nginx pods.

And even if in NodePort mode the LB will target all nodes (and there will be unhealthy endpoints in the LB, which make monitoring unclean or difficult).

acjohnson:
+1 hoping this will be considered still

hcloud/load_balancers.go
```diff
@@ -87,6 +91,57 @@ func (l *loadBalancers) GetLoadBalancerName(ctx context.Context, clusterName str
...
+		return nil
+	}
+
+	if oma, ok := obj.(metav1.ObjectMetaAccessor); ok {
+		return oma.GetObjectMeta()
+	} else if o, ok := obj.(metav1.Object); ok {
+		return o
+	}
+
+	panic(fmt.Errorf("Unknown type: %v", reflect.TypeOf(obj)))
```
fhofherr, line 117: Calling panic runs the risk of the cloud controller manager crashing or ending up in an undefined state. It is better to log the error and return a sensible default, e.g. `nil`, instead of panicing. Returning `nil` does have other consequences, which I address in other comments.

hcloud/load_balancers.go
```diff
@@ -87,6 +91,57 @@ func (l *loadBalancers) GetLoadBalancerName(ctx context.Context, clusterName str
...
 }
 
+// MustGetKey creates a key from Kubernetes resource in the format <namespace>/<name>
+func MustGetKey(obj interface{}) string {
+	if obj == nil {
+		return ""
+	}
+
+	o := MustGetObject(obj)
+	return fmt.Sprintf("%s/%s", o.GetNamespace(), o.GetName())
```
fhofherr, line 101: If `MustGetObject` returns nil this may panic if whatever implements `metav1.Object` is not nil safe. You need to check if `o` is `nil`.

hcloud/load_balancers.go
```diff
@@ -87,6 +91,57 @@ func (l *loadBalancers) GetLoadBalancerName(ctx context.Context, clusterName str
...
+
+		filteredNodes := []*v1.Node{}
+		for _, n := range nodes {
+			if nodeSelector.Matches(labels.Set(n.Labels)) {
+				filteredNodes = append(filteredNodes, n)
+			}
+		}
+
+		return filteredNodes
+
```
fhofherr, line 138: Please remove the empty line here.

hcloud/load_balancers.go
```diff
@@ -87,6 +91,57 @@ func (l *loadBalancers) GetLoadBalancerName(ctx context.Context, clusterName str
...
+	} else {
+		return nodes
+	}
+
```
fhofherr, line 142: Please remove the empty line here.

hcloud/load_balancers.go
```diff
@@ -87,6 +91,57 @@ func (l *loadBalancers) GetLoadBalancerName(ctx context.Context, clusterName str
...
+func MustGetObject(obj interface{}) metav1.Object {
+	if obj == nil {
+		return nil
+	}
+
+	if oma, ok := obj.(metav1.ObjectMetaAccessor); ok {
+		return oma.GetObjectMeta()
+	} else if o, ok := obj.(metav1.Object); ok {
+		return o
+	}
```
fhofherr, line 115: Can you change this to a [type switch](https://go.dev/tour/methods/16) please?

fhofherr (commented):
Hi @steigr, 

thank you for your contribution. 

I've left a few comments in the PR. Additionally the pull request lacks unit tests. Could you please add some for the methods you have created?

acjohnson:
@steigr any further comments on this?

PR #250, opened 2021-10-21
deploy, hcloud.go: update version pointers

flokli:
These still refer to v1.9.0, and haven't been updated to the latest version.

### DIFF
--- a/deploy/ccm-networks.yaml
+++ b/deploy/ccm-networks.yaml
@@ -56,7 +56,7 @@ spec:
           effect: "NoSchedule"
       hostNetwork: true
       containers:
-        - image: hetznercloud/hcloud-cloud-controller-manager:v1.9.1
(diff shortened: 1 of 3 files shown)


flokli:
Ah, spotted the yaml files attached to the releases.

Maybe adding a comment to the top of the file in the repo, about those being deprecated, so new people don't run into this?


`hcloud.go` probably isn't deprecated ;-) The version pointer there should be updated.

LKaemmerling:
@flokli all releases can be found here: https://github.com/hetznercloud/hcloud-cloud-controller-manager#with-networks-support

and the version is automatically overwritten on building the release https://github.com/hetznercloud/hcloud-cloud-controller-manager/blob/master/.goreleaser.yml#L17, that's why there is still the old one :)

flokli:
Ah, makes sense. Could we change the constant in code to something that makes this more obvious (`v0.0.0` maybe)?

PR #251, opened 2021-10-21
deploy: add missing operator: Exists

flokli:
The default value for `operator` is `Equal`, but this doesn't provide
any value to check against, causing this to not be scheduled on control
plane nodes with the following taint:

```
  - effect: NoSchedule
    key: node-role.kubernetes.io/control-plane
    value: "true"
```

Setting the operator to `Exists` will cause this to be scheduled.


LKaemmerling:
Thank you!

PR #290, opened 2022-06-21
Update k8s dependencies to v0.24.2

hakman:
Kubernetes 1.24 bring a change that fixes external CCM preference for user passed node IP to the kubelet (via `--node-ip`).
There should be no issues with backwards compatibility with older Kubernetes versions (tested with 1.23).

xRef: https://github.com/kubernetes/kubernetes/pull/107750


hakman:
Already done as part of #295. Thanks @LKaemmerling!

PR #304, opened 2022-08-03
Flag to disable network routes

maksim-paskal:
allow users to disable networks routes (enabled by default), this can be useful if user use flannel with internal network, managing routes in this scenario does not needed.

disabling network routes also reduce calls to Hetzner API

Signed-off-by: Maksim Paskal <paskal.maksim@gmail.com>


LKaemmerling (approved):
Thank you @maksim-paskal!

PR #309, opened 2022-09-05
Raise cache reload timeout limit

4ND3R50N:
To update the internal cache, the provider is getting all servers of the corresponding project / user. For that, the /servers endpoint is called with page max page size of 50 per call.

But there is an internal timeout of 5s for that calls. And right now, if there is a user with more than 50+ server, one call need more time than 5s.

As a quick fix, we raise the timeout. Internally, we get enough time to investigate more consistent optimizations in API or in the controller manager cache handling.

### DIFF
--- a/internal/hcops/server.go
+++ b/internal/hcops/server.go
@@ -85,7 +85,7 @@ func (c *AllServersCache) getCache(getSrv func() (*hcloud.Server, bool)) (*hclou
 	// Reload from the backend API if we didn't find srv.
 	to := c.LoadTimeout
 	if to == 0 {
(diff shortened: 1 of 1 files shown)


LKaemmerling (approved):

LKaemmerling:
Greetings from 🇮🇹 looks fine but please add a comment why you changed it

4ND3R50N:
Greetings back from DE :-)

I updated the PR description! Have a nice holiday!

mmpetarpeshev:
Thats great , could someone say when this will be released ?

PR #324, opened 2022-10-28
fix: allow to provide tls port for health checks

pmdroid:
The idea of that pull request is that we can assign a health check for the LB with TLS enabled and another without TLS enabled. When doing this i didn't found any solution to that problem besides adding this feature to the cloud controller. 
This currently fixes the problem for us, but it might also be nice to add a list of ports for which health check to enable TLS like: "443,8443"

### DIFF
--- a/internal/annotation/load_balancer.go
+++ b/internal/annotation/load_balancer.go
@@ -186,6 +186,10 @@ const (
 	// performing the health check.
 	LBSvcHealthCheckHTTPPath Name = "load-balancer.hetzner.cloud/health-check-http-path"
 
+	// LBSvcHealthCheckTLSPort specifies the port for which to enable TLS
+	// performing the health check.
+	LBSvcHealthCheckTLSPort Name = "load-balancer.hetzner.cloud/health-check-tls-port"
+
 	// LBSvcHealthCheckHTTPValidateCertificate specifies whether the health
 	// check should validate the SSL certificate that comes from the target
 	// nodes.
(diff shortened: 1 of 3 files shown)


internal/annotation/load_balancer.go
```diff
@@ -186,6 +186,10 @@ const (
 	// performing the health check.
 	LBSvcHealthCheckHTTPPath Name = "load-balancer.hetzner.cloud/health-check-http-path"
 
+	// LBSvcHealthCheckTLSPort specifies the port for which to enable TLS
+	// performing the health check.
+	LBSvcHealthCheckTLSPort Name = "load-balancer.hetzner.cloud/health-check-tls-port"
```
apricote, line 191: What do you think about naming this similarly to `LBSvcHealthCheckHTTPValidateCertificate`/`load-balancer.hetzner.cloud/health-check-http-validate-certificate`? These options basically do the same, just that the new option is restricting to a single port.

internal/hcops/load_balancer.go
```diff
@@ -1069,6 +1068,20 @@ func (b *hclbServiceOptsBuilder) extractHealthCheck() {
 		return nil
 	})
 
+	b.do(func() error {
+		port, err := annotation.LBSvcHealthCheckTLSPort.IntFromService(b.Service)
```
apricote, line 1072: You could use `IntsFromService()` to get an `[]int` back for allowing multiple ports.

internal/hcops/load_balancer.go
```diff
@@ -1069,6 +1068,20 @@ func (b *hclbServiceOptsBuilder) extractHealthCheck() {
...
+		if errors.Is(err, annotation.ErrNotSet) {
+			return nil
+		}
+		if err != nil {
+			return fmt.Errorf("%s: %w", op, err)
+		}
+		if b.listenPort == port {
+			b.healthCheckOpts.httpOpts.TLS = hcloud.Bool(true)
```
apricote, line 1080:
```suggestion
			b.healthCheckOpts.httpOpts.TLS = hcloud.Ptr(true)
```

apricote (commented):
Hey @pmdroid,

sorry for letting this go stale so many times. I have no taken a look at the code and have some suggestions. I can understand if you no longer have to time/interest to update the PR, in that case you can just let me know and I can take it over :)

PR #344, opened 2023-02-22
fix(deploy): node.kubernetes.io/not-ready taint is NoExecute

flokli:
… not NoSchedule

See
https://github.com/kubernetes/kubernetes/blob/4dd887797fe53816faf3cd4d65d600a606714f5d/pkg/controller/nodelifecycle/node_lifecycle_controller.go#L77


apricote (approved):

apricote:
Thanks for the contribution @flokli!

PR #375, opened 2023-03-01
chore: initial basic helm chart

samcday:
chart templates created by running the following:

```
 $ tail -n+2 deploy/ccm.yaml | kubectl slice -t '{{.kind | lower}}.yaml' -o chart/templates/
```

`generate-deployment-yaml.sh` is updated to then do the previous thing in reverse. That is, it runs `helm template chart` and the contents go in deploy/ccm.yaml

As a result of this change, deploy/ccm.yaml.tmpl is no longer needed.

In follow-up commits the dev and network flavors will be generated from the new chart, and the corresponding .tmpl files will go away.

### DIFF
--- a/scripts/generate-deployment-yamls.sh
+++ b/scripts/generate-deployment-yamls.sh
@@ -9,6 +9,9 @@ if [[ -z $VERSION ]]; then
     exit 1
 fi
 
+cat chart/Chart.yaml | sed -e "s/version: .*/version: $VERSION/" > chart/Chart.yaml.new && mv chart/Chart.yaml{.new,}
+helm template chart > deploy/ccm.yaml
+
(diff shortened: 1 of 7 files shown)


scripts/generate-deployment-yamls.sh
```diff
@@ -9,6 +9,9 @@ if [[ -z $VERSION ]]; then
     exit 1
 fi
 
+cat chart/Chart.yaml | sed -e "s/version: .*/version: $VERSION/" > chart/Chart.yaml.new && mv chart/Chart.yaml{.new,}
```
apricote, line 12: I think we should use `release-please` for this in the future, it can update versions in any file.

apricote (approved):

PR #380, opened 2023-03-03
refactor(e2e): remove dev-ccm manifests

samcday:


apricote (approved):
Id personally use the scope `refactor(e2e)` for this instead of `chore`

Issue #384, opened 2023-03-06
Instances(V1) interface access is not cached, API rate limited


hifi:
We run a fairly large cluster on Hetzner. Initially based on errors I assumed we had hit #330 but it wasn't as much API access than I expected.

Debugging with `HCLOUD_DEBUG` enabled I saw that about each five minutes `/v1/servers/{id}` was being called by the controller twice for each node. This became a problem when the node count exceeded the default API regeneration rate (1 req/s) at around 150 nodes.

I think it would be reasonable to use and refresh a shared `hcops.AllServersCache` as it uses the paginated call to build a full cache and it will drop the number of API calls down significantly without functionality loss or making the cache time longer.

hifi:
I hacked around with `hcops.AllServersCache` here: https://github.com/beeper/hcloud-cloud-controller-manager/commit/733bcf9e3e2745ba85374894915bbefb90649a60

Based on the results this cut down the requests from ~200 on startup and constantly draining because of #330 to 4 per minute max when the cache expires.

Because the paginated API for servers isn't transactional and would likely give incorrect results if servers are being added/removed during pagination a good implementation that wouldn't drain the API rate limit nor cause unnecessary cache misses would be better as a pull-through cache that is initialized from the full list on startup.

So if the server cache is seeded with the paginated data but refreshed based on access (and per-instance expiry time) it would prevent the API flood on startup and also always be up-to-date. Because of #330 it would also need to have an increasing backoff time to fetch a missing server from the API to prevent it from draining API rate limit by making the backoff long enough that even a large amount of out-of-cloud servers wouldn't cause that.

What do you think of this approach and this issue in general?

apricote:
Hey @hifi,

right now you should be able to set [`--node-status-update-frequency`](https://github.com/kubernetes/cloud-provider/blob/4162b590014735c81ac6db5e3fbf7edaae3d4706/options/options.go#L137) to reduce the frequency of node status updates. This defaults to the [5 minutes](https://github.com/kubernetes/cloud-provider/blob/4162b590014735c81ac6db5e3fbf7edaae3d4706/config/v1alpha1/defaults.go#L37) you mentioned.

Before we add any complexity to the existing interface, I would prefer to migrate our usage from `Instances` to [`InstancesV2`](https://github.com/kubernetes/cloud-provider/blob/4162b590014735c81ac6db5e3fbf7edaae3d4706/cloud.go#L52-L56):

```go
[...]

PR #386, opened 2023-03-09
refactor: unnecessary API call in instance reconciliation

apricote:
Calls to `InstanceV2.InstanceMetadata()` caused 2 API calls, one to retrieve the server and the second to check if the configured network exists.

The second call is not necessary because we already know all networks that the server is attached to exist. By just comparing the `nameOrID` locally we can achieve the same result.

Related to #384


samcday (approved):
gr8! :+1:

apricote:
The network name is not being returned in the `GET v1/servers/:id` call, so I had to pass the network ID from `newCloud()`. This is cleaner anyway, because now we read the environment variable only in one location.

PR #387, opened 2023-03-09
fix: self-reported version not correct

apricote:
The release process was changed after v1.9.1 to use `ldflags` with `-X` to dynamically set the version in the binary.

 This silently failed because with `-X` you can only change the value of  _variables_, not _constants_ but our `providerVersion` was a constant beforehand.

 I also changed the default value to "unknown" to better reflect the value if the binary was not properly built.


samcday (approved):
:see_no_evil:

samcday (approved):

PR #392, opened 2023-03-14
fix(ci): wrong version published when two tags point to same commit 

apricote:
The tags v1.14.0-rc.2 and v1.14.0 point to the same commit. Goreleaser always releases v1.14.0-rc.2 and we don't get a proper v1.14.0 release.

By specifying the release that goreleaser should handle, we can rely on Github to specify the correct tag for the current workflow.


samcday (approved):
:thinking:

PR #394, opened 2023-03-14
chore: multiple improvements to the release process

apricote:
### chore: fix release missing deployment files

The path for the deployment files was changed in https://github.com/hetznercloud/hcloud-cloud-controller-manager/pull/375 but not updated here. This cause the v1.14.1 release to be missing the release artifacts until I manually uploaded them.

### chore: no draft releases

In the current release process we need to "publish" the release twice, once to create the tag and trigger `goreleaser`, and then a second time because goreleaser turned the release back into a draft.

We can just skip the drafting step and let goreleaser publish directly.

### chore: simplify chart version update

Simplify the command used to update the Chart version and add some comments.

### DIFF
--- a/scripts/generate-deployment-yamls.sh
+++ b/scripts/generate-deployment-yamls.sh
@@ -9,8 +9,12 @@ if [[ -z $VERSION ]]; then
     exit 1
 fi
 
-cat chart/Chart.yaml | sed -e "s/version: .*/version: $VERSION/" > chart/Chart.yaml.new && mv chart/Chart.yaml{.new,}
+# Update version
+sed -e "s/version: .*/version: $VERSION/" --in-place chart/Chart.yaml
+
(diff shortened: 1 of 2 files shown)


scripts/generate-deployment-yamls.sh
```diff
@@ -9,8 +9,12 @@ if [[ -z $VERSION ]]; then
     exit 1
 fi
 
-cat chart/Chart.yaml | sed -e "s/version: .*/version: $VERSION/" > chart/Chart.yaml.new && mv chart/Chart.yaml{.new,}
+# Update version
+sed -e "s/version: .*/version: $VERSION/" --in-place chart/Chart.yaml
```
samcday, line 13: I originally avoided this because I think now this won't work on BSD sed xD

samcday (approved):

PR #399, opened 2023-03-15
feat(chart): metrics + PodMonitor support

samcday:


chart/templates/deployment.yaml
```diff
@@ -8,11 +8,11 @@ spec:
   revisionHistoryLimit: 2
   selector:
     matchLabels:
-      app: hcloud-cloud-controller-manager
+      {{- include "hcloud-cloud-controller-manager.selectorLabels" . | nindent 6 }}
```
apricote, line 11: The selector is [immutable](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#selector):

> Also note that .spec.selector is immutable after creation of the Deployment in apps/v1.

I think we need to keep `app` in there, or we would need to declare a breaking change, as users would need to re-create the deployment.

chart/templates/_helpers.tpl
```diff
@@ -0,0 +1,8 @@
+{{- define "hcloud-cloud-controller-manager.name" -}}
+{{- default $.Chart.Name $.Values.nameOverride | trunc 63 | trimSuffix "-" }}
```
apricote, line 2: The [Helm Docs](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-default-function) have no example for passing 2 arguments to `default` and instead suggest:

```suggestion
{{- $.Values.nameOverride | default $.Chart.Name | trunc 63 | trimSuffix "-" }}
```

chart/templates/deployment.yaml
samcday: We can pretend that Helm charts aren't stable versioned artifacts, right? Since we haven't documented them yet and our Helm repo has a big scary warning that it's not ready yet (and it's not!).

If so, then I think the cleanest workaround to this is to make the selectorLabels configurable in values.yaml. The static (legacy-ish) manifests we generate can continue to use the old under-specified `app` label, whilst usage of the Helm chart directly defaults to the "modern" (using the standard labels + distinguishing multiple releases) labels.

WDYT?

apricote: Sounds good :+1:

chart/templates/_helpers.tpl
samcday: I mean I don't really care, but both styles are valid in this case, it seems. The version you see currently is what you get when you run `helm create barf`.

apricote: Okay, i am fine with either version, I was just worried that it might be invalid syntax

chart/templates/deployment.yaml
samcday: Done. Re-ran `generate-deployment-yamls.sh` and as you can see it reshuffled things a bit (and sneakily introduced the metrics container port) but the selector labels has been left alone.

apricote (approved):

PR #402, opened 2023-03-17
fix: handle nil servers in InstanceV2 #398

### LINKED ISSUE #398: Issues if lookupServer() returns nil server
As far as I understand, the lookupServer function can return a nil server. You also indicate that in the comment above (even though it seems as if the comment is missing its last part https://github.com/hetznercloud/hcloud-cloud-controller-manager/blob/main/hcloud/instances.go#L49).

However, are the functions InstanceShutdown and InstanceMetadata really able to handle a nil server? I think they would panic.

I guess that an error should be returned if the server is not found, i.e. nil, and the InstanceShutdown or InstanceMetadata functions are called. What do you think?

apricote:
We did not properly handle nil pointers returned from `lookupServer`.

Based on the current usage in `k/cloud-provider` the methods are called in these contexts:

- `InstanceShutdown` is only called for instances that actually do exist (as verified by a prior call to `InstanceExists`)
- `InstanceMetadata` is called without verifying that the instance exists, one call site accepts a nil InstanceMetadata, but the other does not.
Closes #398

### DIFF
--- a/hcloud/instances.go
+++ b/hcloud/instances.go
@@ -46,7 +46,7 @@ func newInstances(client *hcloud.Client, addressFamily addressFamily, networkID 
 
 // lookupServer attempts to locate the corresponding hcloud.Server for a given v1.Node
 // It returns an error if the Node has an invalid provider ID or if API requests failed.
-// It can return a nil hcloud.Server if no
+// It can return a nil [*hcloud.Server] if neither the ProviderID nor the Name matches an existing server.
 func (i *instances) lookupServer(ctx context.Context, node *v1.Node) (*hcloud.Server, error) {
 	var server *hcloud.Server
 	if node.Spec.ProviderID != "" {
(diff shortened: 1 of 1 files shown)


hcloud/instances.go
```diff
@@ -46,7 +46,7 @@ func newInstances(client *hcloud.Client, addressFamily addressFamily, networkID
 
 // lookupServer attempts to locate the corresponding hcloud.Server for a given v1.Node
 // It returns an error if the Node has an invalid provider ID or if API requests failed.
-// It can return a nil hcloud.Server if no
+// It can return a nil [*hcloud.Server] if neither the [ProviderID] nor the Name matches an existing server.
```
samcday, line 49: What does the `[ProviderID]` syntax here do? Shouldn't it be something like `node.Spec.ProviderID`? And if that does some kind of cool symbol linking in the generated docs, do we also want that for `node.Spec.Name`?

samcday (approved):

hcloud/instances.go
apricote: Thats... not doing anything, I tried to set it to `[v1.Node.Spec.ProviderID]` but that did not work, and then forgot to remove the braces. The linking unfortunately only works for top level symbols.

PR #407, opened 2023-03-21
chore(chart): basic README

samcday:


apricote (commented):
Do you reckon we should include a migration guide (e.g. delete old deployment) for users who want to switch to the helm chart in this PR?

apricote (approved):

PR #408, opened 2023-03-21
refactor: Update & Fix golangci-lint

apricote:
Update the `golangci-lint` configuration to replace any deprecated linters and fix issues that are present. This also includes issues that were present before this PR, as the linter was not running in our CI and quite a few issues creeped in over time.

This also consolidates all the CI workflows running during `push` & `pull_requests` into a single workflow with multiple jobs, as I personally find that easier to navigate in the web interface.


apricote:
Seems like the failure before the rebase was a flake (across all runs?). Merging..

PR #417, opened 2023-04-06
feat(goreleaser): produce OCI manifest images

samcday:


apricote (approved):
Test failures are flakes.

PR #437, opened 2023-05-04
ci: setup release-please

### LINKED ISSUE #431: Update image tag in deployment
The deployment https://github.com/hetznercloud/hcloud-cloud-controller-manager/blob/v1.15.0/deploy/ccm-networks.yaml still has the old image tag 1.13.2. Can you please update it to 1.15.0 or later?

I am using this with my cluster deployment.

apricote:
Setup release-please to automate creating new tags according to semver rules.

Updated referenced versions to latest released version to fix #431, new version should be bumped automatically by release-please.

Backfill changelog entries that were only on GitHub Releases.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -1,0 +1,20 @@
+on:
+  push:
+    branches:
+      - main
+name: release-please
+
+jobs:
+  release-please:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: google-github-actions/release-please-action@v3
+        with:
+          release-type: go
+          package-name: hcloud-cloud-controller-manager
+
+          # These files reference the version in the OCI image tag.
+          extra-files: |
+            deploy/ccm.yaml
+            deploy/ccm-networks.yaml
+            chart/Chart.yaml
--- a/deploy/ccm-networks.yaml
+++ b/deploy/ccm-networks.yaml
@@ -83,7 +83,7 @@ spec:
                 secretKeyRef:
                   name: hcloud
                   key: network
-          image: hetznercloud/hcloud-cloud-controller-manager:v1.13.2
+          image: hetznercloud/hcloud-cloud-controller-manager:v1.15.0 # x-release-please-version
           ports:
             - name: metrics
               containerPort: 8233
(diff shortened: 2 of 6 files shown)


.github/workflows/release-please.yml
```diff
@@ -0,0 +1,20 @@
+on:
+  push:
+    branches:
+      - main
```
samcday, line 4: Should this also be triggering on `tags: {}`?

deploy/ccm-networks.yaml
```diff
@@ -83,7 +83,7 @@ spec:
                 secretKeyRef:
                   name: hcloud
                   key: network
-          image: hetznercloud/hcloud-cloud-controller-manager:v1.13.2
+          image: hetznercloud/hcloud-cloud-controller-manager:v1.15.0 # x-release-please-version
```
samcday, line 86: Did you manually add this? If so I think it will get overwritten on the next run of `script/barf-deploy-manifests-thingie.sh`, right?

samcday (approved):

.github/workflows/release-please.yml
apricote: Nope, this workflow is just for creating the tag, once the tag is created, the usual release pipeline is used.

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,20 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: google-github-actions/release-please-action@v3
+        with:
+          release-type: go
+          package-name: hcloud-cloud-controller-manager
+
+          # These files reference the version in the OCI image tag.
+          extra-files: |
+            deploy/ccm.yaml
```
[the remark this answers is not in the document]
apricote: We have no step that actually renders and commits the updated manifests right now. That only happens on release and then the updated manifests are only pushed as release assets, not into the repo.

For users that reference these manifests via kustomize this breaks, see #431

deploy/ccm-networks.yaml
apricote: Good catch! Added it to the helm template so its included when we regenerate the templates.

apricote:
I had to add the BEGIN_COMMIT_OVERRIDE to this PR to get the proper release version. The tag for `1.15.0` is not on the `main` branch and was therefore not considered by release-please, resulting in a tag for `1.16.0-rc.0` by default.

PR #444, opened 2023-05-09
ci: fix release assets not available

apricote:
The default token GH_TOKEN in workflows does not trigger new events. We rely on the "tag" event created by release-please to build our release assets.

By using a token for "hcloud-bot", the event will be properly created.

We explicitly only run the workflow job if the repository is the main repo, as otherwise this workflow will fail in forks because the secret is not available.

See also hetznercloud/cli#476

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -6,10 +6,14 @@ name: release-please
 
 jobs:
   release-please:
+    # The secret HCLOUD_BOT_TOKEN is only available on the main repo, not in forks.
+    if: github.repository == 'hetznercloud/hcloud-cloud-controller-manager'
+
     runs-on: ubuntu-latest
     steps:
       - uses: google-github-actions/release-please-action@v3
         with:
+          token: ${{ secrets.HCLOUD_BOT_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release-please.yml
```diff
@@ -6,10 +6,14 @@ name: release-please
 
 jobs:
   release-please:
+    # The secret HCLOUD_BOT_TOKEN is only available on the main repo, not in forks.
+    if: github.repository == 'hetznercloud/hcloud-cloud-controller-manager'
```
samcday, line 10: Mm, we could also detect if the secret is set here, right? Not sure how hard we should work here to ensure releases can be cut for forks, though.

samcday (approved):

.github/workflows/release-please.yml
apricote: > Not sure how hard we should work here to ensure releases can be cut for forks, though.

Not at all in my opinion, thats why I just deactived the workflow for forks.

If left running for forks, this will create notifications of failing workflows for people having forks, i encountered this issue with the cluster-api-provider-hetzner.

PR #448, opened 2023-05-12
chore(ci): fix go version in goreleaser workflow

samcday:
CI is currently [installing Go 1.2](https://github.com/hetznercloud/hcloud-cloud-controller-manager/actions/runs/4958991607/jobs/8872533527), which incidentally does not have `go mod` support 🙈


apricote (approved):

samcday:
YAML type coercion rules are a delight, as always

apricote:
> YAML type coercion rules are a delight, as always

A marvel that we never had to deal with Go 1.19000000001

PR #458, opened 2023-06-16
feat(helm): allow to manually set network name or ID

pitwegner:


chart/Chart.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: v2
 name: hcloud-cloud-controller-manager
 type: application
-version: v1.15.0 # x-release-please-version
+version: v1.16.0 # x-release-please-version
```
apricote, line 4: Please remove this change. The chart version will be updated automatically for new releases. You can check out #439 for an example of this.

chart/values.yaml
```diff
@@ -49,6 +49,11 @@ networking:
   # If networking is enabled, clusterCIDR must match the PodCIDR subnet your cluster has been configured with.
   # The default "10.244.0.0/16" assumes you're using Flannel with default configuration.
   clusterCIDR: 10.244.0.0/16
+  networkId:
```
apricote, line 52: This can be the name or id, having "Id" in the field name is misleading IMO. What do you think about naming this just "network"?

---

If a user were to use `value: network-name`, how would they set the value so the `valueFrom` is removed?

apricote (changes requested):
Looking good in general @pitwegner :) Thanks for the contribution! I have some small requests before we can merge this.

chart/Chart.yaml
pitwegner: done

chart/values.yaml
pitwegner: removed the id. Regarding the value... How would someone do it in case of the HCLOUD_TOKEN env var? You'd need to prefer `value` over `valueFrom` using a control sequence. I could add that if you like?

apricote:
Ah yes, our e2e tests are currently broken for PRs from forks. Sorry about that. I will run them locally and report back :)

apricote (approved):
E2E tests were successful :)

apricote:
Oh yea, linting fails because of #459.

PR #459, opened 2023-06-16
ci: fix linting issues

apricote:
A new version of golangci-lint updated depguard, which in turn started returning errors because it was missing a config. Previously depguard did not do anything as we never configured a list of allowed/denied imports.

In general, we already have gomodguard for linting imports.

To avoid sudden breakages of our CI in the future, I pinned the exact version of golangci-lint to the current release. Upgrading this should happen through a PR process (e.g. by renovate) so we can notice new linting failures and fix them before merging to main.


DrBu7cher (approved):
:+1:

PR #462, opened 2023-06-28
fix(helm-chart): resource namespace and name

simonostendorf:
This should fix the issue that you can change the release namespace with helm, but the deployment will not take care of it. 

Also this allows the user to use the nameoverwrite that was already defined inside the chart helpers.

### DIFF
--- a/chart/templates/clusterrolebinding.yaml
+++ b/chart/templates/clusterrolebinding.yaml
@@ -8,5 +8,5 @@ roleRef:
   name: cluster-admin
 subjects:
   - kind: ServiceAccount
-    name: cloud-controller-manager
-    namespace: kube-system
+    name: {{ include "hcloud-cloud-controller-manager.name" . }}
+    namespace: {{ .Release.Namespace }}
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -1,8 +1,8 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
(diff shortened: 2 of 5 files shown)


apricote (changes requested):
Nice! Could you update the field in the [ClusterRoleBinding](https://github.com/hetznercloud/hcloud-cloud-controller-manager/blob/185a5c0080bcd5403070f702e93cae2247cff120/chart/templates/clusterrolebinding.yaml#L11-L12) to match the new name(+space) of the ServiceAccount?

apricote (approved):

simonostendorf:
> Nice! Could you update the field in the [ClusterRoleBinding](https://github.com/hetznercloud/hcloud-cloud-controller-manager/blob/185a5c0080bcd5403070f702e93cae2247cff120/chart/templates/clusterrolebinding.yaml#L11-L12) to match the new name(+space) of the ServiceAccount?

Sorry i forgot to change it there.

apricote:
Sorry for the delay. Our e2e tests are currently broken for PRs from forks and I wanted to run them locally. I found that in the e2e tests we do not set the Helm Chart namespace which caused the pods to run in `default`. The required secret was added to `kube-system` though. I [added the namespace](https://github.com/hetznercloud/hcloud-cloud-controller-manager/pull/462/commits/26319b9b87af40c4548578f54c890f46fb96dbb7) for our E2E tests and they succeeded locally.

apricote (approved):

PR #471, opened 2023-07-18
feat(helm): allow to set labels and annotations for podMonitor

pitwegner:
This PR adds the ability to set labels and annotations to the podMonitor. Setting the label is usually necessary for the podMonitor to be picked up by the Prometheus Operator (https://github.com/prometheus-operator/prometheus-operator/issues/3636#issuecomment-796688902). Setting an annotation helps to defer the deployment of the custom resource if e.g. deploying with ArgoCD (https://argo-cd.readthedocs.io/en/stable/user-guide/sync-options/#skip-dry-run-for-new-custom-resources-types).

### DIFF
--- a/chart/templates/podmonitor.yaml
+++ b/chart/templates/podmonitor.yaml
@@ -5,6 +5,14 @@ kind: PodMonitor
 metadata:
   name: {{ include "hcloud-cloud-controller-manager.name" . }}
   namespace: {{ .Release.Namespace }}
+  labels:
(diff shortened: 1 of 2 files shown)


apricote (approved):

apricote:
E2e tests are broken for PRs from forks right now. Tested this locally and it works (as expected, e2e test does not use `PodMonitor`).

apricote:
Thanks a lot for the contribution @pitwegner!

PR #478, opened 2023-07-19
ci: check deployment manifests

apricote:
We had some issues where the committed files in deploy/ did not match the files included with releases. This job makes sure that the files in deploy/ always match the current output from running "helm chart template".

[Example pipeline that demonstrates the output](https://github.com/hetznercloud/hcloud-cloud-controller-manager/actions/runs/5597677338)

Related to #476.


jooola:
This looks like a job for pre-commit :wink:

Looks good to me.

PR #498, opened 2023-08-21
feat(charts): add replica count

simonostendorf:
This adds configurable replicas to the helm chart.

### DIFF
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -50,8 +50,8 @@ spec:
             - "--allocate-node-cidrs=true"
             - "--cluster-cidr={{ $.Values.networking.clusterCIDR }}"
             {{- end }}
-            {{- if (gt (int $.Values.replicaCount) 1) }}
-            - "--leader-elect=true"
+            {{- if (eq (int $.Values.replicaCount) 1) }}
+            - "--leader-elect=false"
             {{- end }}
           env:
             {{- range $key, $value := $.Values.env }}
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -4,7 +4,7 @@ metadata:
   name: {{ include "hcloud-cloud-controller-manager.name" . }}
   namespace: {{ .Release.Namespace }}
 spec:
-  replicas: 1
+  replicas: {{ .Values.replicaCount }}
   revisionHistoryLimit: 2
   selector:
     matchLabels:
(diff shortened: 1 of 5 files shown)


apricote (changes requested):
In case the user wants to deploy more than 1 replica, we also need to enable leader election so only one of the pods is actively processing at the same time.

I would suggest we add something like this to the container command:

```helm
{{- if (gt $.Values.replicaCount 1) }}
- "--leader-elect=true"
{{ -end }}
```

And document that the flag is set automatically in `values.args`  and that setting `values.replicaCount > 1` enables leader election.

simonostendorf:
> In case the user wants to deploy more than 1 replica, we also need to enable leader election so only one of the pods is actively processing at the same time.
> 
> I would suggest we add something like this to the container command:
> 
> ```
> {{- if (gt $.Values.replicaCount 1) }}
> - "--leader-elect=true"
> {{ -end }}
> ```
> 
> And document that the flag is set automatically in `values.args` and that setting `values.replicaCount > 1` enables leader election.

Thanks for your comment @apricote 
I think the automatic leader election is a good idea. I have implemented it as requested.

chart/templates/deployment.yaml
```diff
@@ -50,6 +50,9 @@ spec:
             - "--allocate-node-cidrs=true"
             - "--cluster-cidr={{ $.Values.networking.clusterCIDR }}"
             {{- end }}
+            {{- if (gt $.Values.replicaCount 1) }}
```
apricote, line 53: Freestyling helm templates without testing it never works, this one is on me, sorry.

```suggestion
            {{- if (gt (int $.Values.replicaCount) 1) }}
```

Current version fails e2e tests with:

```
std out err: %!(EXTRA *errors.errorString=Error: template: hcloud-cloud-controller-manager/templates/deployment.yaml:53:20: executing "hcloud-cloud-controller-manager/templates/deployment.yaml" at <gt $.Values.replicaCount 1>: error calling gt: incompatible types for comparison
```

apricote (changes requested):

apricote:
Oh and also, can you run `scripts/update-deployment-yamls.sh` and commit the changes in `deploy/`?

chart/templates/deployment.yaml
simonostendorf: implemented, sorry i didn't saw the error

simonostendorf:
> Oh and also, can you run `scripts/update-deployment-yamls.sh` and commit the changes in `deploy/`?

done

apricote (commented):
Sorry :/

apricote (approved):
Thank you very much for the contribution :)

Issue #506, opened 2023-09-08
LoadBalancer targets too many nodes


lloesche:
Hi there, I feel like I'm doing something very obvious wrong but I can't seem to figure out what it is and I'm unsure if I messed up the config or if the cloud controller is having some snafu.

I deployed an ingress-nginx as a Deployment with 3 replicas. The LoadBalancer Service object is created with the following `kubectl -n ingress-nginx get service/ingress-nginx-controller -o yaml` output. Sorry for the long object, I didn't want to leave anything out, but the interesting sections are likely annotations and spec.selector:

```yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    load-balancer.hetzner.cloud/hostname: lb.eu.fixcloud.io
    load-balancer.hetzner.cloud/http-redirect-https: "true"
    load-balancer.hetzner.cloud/location: nbg1
    load-balancer.hetzner.cloud/name: fix-ingress
    load-balancer.hetzner.cloud/use-private-ip: "true"
    load-balancer.hetzner.cloud/uses-proxyprotocol: "true"
    meta.helm.sh/release-name: ingress-nginx
    meta.helm.sh/release-namespace: ingress-nginx
  creationTimestamp: "2023-09-08T17:54:47Z"
  finalizers:
  - service.kubernetes.io/load-balancer-cleanup
  labels:
    app.kubernetes.io/component: controller
    app.kubernetes.io/instance: ingress-nginx
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: ingress-nginx
    app.kubernetes.io/part-of: ingress-nginx
    app.kubernetes.io/version: 1.8.1
    helm.sh/chart: ingress-nginx-4.7.1
  name: ingress-nginx-controller
  namespace: ingress-nginx
  resourceVersion: "4928"
  uid: 196caba9-f30f-4adc-bbfe-3b3107322611
spec:
  allocateLoadBalancerNodePorts: true
  clusterIP: 10.43.7.184
  clusterIPs:
  - 10.43.7.184
  externalTrafficPolicy: Cluster
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - appProtocol: http
    name: http
    nodePort: 32675
    port: 80
    protocol: TCP
    targetPort: http
  - appProtocol: https
    name: https
    nodePort: 31185
    port: 443
    protocol: TCP
[...]

Issue #373, opened 2023-03-01
Target nodes in a specific region only


renchap:
We have a multi region cluster (HEL and FSN), as we have apps deployed in both of those (but each app is contained into one region only).

We deployed 2 `ingress-nginx` instances, one in each region, using node selectors so it only runs in its own region. Then each app configures its `IngressClassName` to be handled by the regional nginx instances.

This created 2 Hetzner load-balancers, one in each region, but targets for those loadbalancers contain every node in the cluster. It would be really useful to be able to add an annotation to restrict targets to the correct region, as we currently can have trafic hitting our `eu-central` loadbalancer, then the data is sent to a node in HEL1, and it is routed back to FSN1 as the nginx pods are only running there.

apricote:
Hey @renchap, this is not supported right now, there is only the possibility to exclude a node from all LoadBalancers using label [`node.kubernetes.io/exclude-from-external-load-balancer`](https://kubernetes.io/docs/reference/labels-annotations-taints/#node-kubernetes-io-exclude-from-external-load-balancer).

renchap:
Yes, I am using this for our controlplane nodes, but it would prevent any nginx instances to be scheduled here, where I want some of them to be.

Would you be open to a PR implementing this? Do you have a preference in the way it is done (label name…)

apricote:
We are open to a PR that implements this. I would prefer the following implementation. Please comment if you think something else is is more appropriate :)

#### Annotation

We add a new annotation:

```go
LBNodeSelector Name = "load-balancer.hetzner.cloud/node-selector"
```

This annotation contains a label selector string matching the [string based format](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#list-and-watch-filtering) used in `kubectl`.

#### Evaluation

The label selector can be parsed and evaluated using

- [`labels.Parse()`](https://pkg.go.dev/k8s.io/apimachinery/pkg/labels#Parse) and
- [`labels.Selector.Matches()`](https://pkg.go.dev/k8s.io/apimachinery/pkg/labels#Selector)

This happens in `github.com/hetznercloud/hcloud-cloud-controller-manager/hcloud#EnsureLoadBalancer` before the nodes are passed to `hcops`/`lbops`.

PR #514, opened 2023-09-14
feat(load-balancer): Add new node-selector annotation

xoxys:
Fixes: https://github.com/hetznercloud/hcloud-cloud-controller-manager/issues/506, https://github.com/hetznercloud/hcloud-cloud-controller-manager/issues/373

### DIFF
--- a/hcloud/load_balancers.go
+++ b/hcloud/load_balancers.go
@@ -55,7 +55,7 @@ func matchNodeSelector(svc *corev1.Service, nodes []*corev1.Node) ([]*corev1.Nod
 	if v, ok := annotation.LBNodeSelector.StringFromService(svc); ok {
 		selector, err = labels.Parse(v)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("unable to parse the node-selector annotation: %w", err)
 		}
 	}
 
--- a/hcloud/load_balancers.go
+++ b/hcloud/load_balancers.go
@@ -130,7 +130,7 @@ func (l *loadBalancers) EnsureLoadBalancer(
 		return nil, fmt.Errorf("%s: %v", op, err)
 	}
 
-	nodeNames := make([]string, len(nodes))
+	nodeNames := make([]string, len(selectedNodes))
 	for i, n := range selectedNodes {
 		nodeNames[i] = n.Name
 	}
@@ -177,7 +177,7 @@ func (l *loadBalancers) EnsureLoadBalancer(
 	}
 	reload = reload || servicesChanged
 
-	targetsChanged, err := l.lbOps.ReconcileHCLBTargets(ctx, lb, svc, nodes)
+	targetsChanged, err := l.lbOps.ReconcileHCLBTargets(ctx, lb, svc, selectedNodes)
 	if err != nil {
 		return nil, fmt.Errorf("%s: %w", op, err)
 	}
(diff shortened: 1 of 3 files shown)


hcloud/load_balancers.go
```diff
@@ -44,6 +45,29 @@ func newLoadBalancers(lbOps LoadBalancerOps, ac hcops.HCloudActionClient, disabl
...
+	var (
+		err           error
+		selectedNodes []*corev1.Node
+	)
+
+	selector := labels.Everything()
+	if v, ok := annotation.LBNodeSelector.StringFromService(svc); ok {
+		selector, err = labels.Parse(v)
+		if err != nil {
+			return nil, err
```
apricote, line 58: An error here is exposed to the user as a `Event`. IMO its hard to understand with the current message where the issue actually happened, and what the user should do to fix this.

Error message from event: `Error syncing load balancer: failed to ensure load balancer: hcloud/loadBalancers.EnsureLoadBalancer: found '==', expected: !, identifier, or 'end of string'`

I would prefer if we added something along the lines of "unable to parse node-selector annotation:" to the error:

```suggestion
			return nil, fmt.Errorf("unable to parse the node-selector annotation: %w", err)
```

hcloud/load_balancers.go
```diff
@@ -97,13 +121,19 @@ func (l *loadBalancers) EnsureLoadBalancer(
...
+		lb            *hcloud.LoadBalancer
+		err           error
+		selectedNodes []*corev1.Node
 	)
 
-	nodeNames := make([]string, len(nodes))
-	for i, n := range nodes {
+	selectedNodes, err = matchNodeSelector(svc, nodes)
+	if err != nil {
+		return nil, fmt.Errorf("%s: %v", op, err)
```
apricote, line 132:
```suggestion
		return nil, fmt.Errorf("%s: %w", op, err)
```

If you wrap errors, consumers can actually unwrap & inspect the the error later on if thats what they want to do.

https://go.dev/blog/go1.13-errors

hcloud/load_balancers.go
```diff
@@ -236,12 +266,18 @@ func (l *loadBalancers) UpdateLoadBalancer(
...
+		lb            *hcloud.LoadBalancer
+		err           error
+		selectedNodes []*corev1.Node
 	)
 
-	nodeNames := make([]string, len(nodes))
-	for i, n := range nodes {
+	selectedNodes, err = matchNodeSelector(svc, nodes)
+	if err != nil {
+		return fmt.Errorf("%s: %v", op, err)
```
apricote, line 276:
```suggestion
		return fmt.Errorf("%s: %w", op, err)
```

internal/annotation/load_balancer.go
```diff
@@ -98,6 +98,10 @@ const (
 	// Mutually exclusive with LBLocation.
 	LBNetworkZone Name = "load-balancer.hetzner.cloud/network-zone"
 
+	// LBNodeSelector specifies a Kubernetes label selector that matches the
+	// string based format. This is used to select the nodes added to the Load Balancer.
```
apricote, line 102:
```suggestion
	// LBNodeSelector can be set to restrict which Nodes are added as targets to the
	// Load Balancer. It accepts a Kubernetes label selector string, using either the
	// set-based or equality-based formats. 
	//
	// If the selector can not be parsed, the targets in the Load Balancer are not
	// updated and an Event is created with the error message.
	//
	// Format: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors
```

apricote (changes requested):
Thank you very much @xoxys! This works very well in my tests, I only have some UX suggestions left :)

apricote:
Dont worry about the failing e2e tests. They are currently broken for external contributors. I ran them locally and everything passed :)

hcloud/load_balancers.go
xoxys: Ah yeah, that makes sense :+1: just copied the snippet and haven't thought enough about it :see_no_evil:

apricote (approved):

PR #516, opened 2023-09-15
feat: build with Go 1.21

apricote:
Also bump golangci-lint to v1.54.2.


jooola (approved):

jooola:
I rarely use it, but this feels appropriate, why not use the build prefix e.g. `build: upgrade to go 1.21` ?

apricote:
> I won't generate a changelog entry though.

Yea, that's why I used `feat` instead

Issue #523, opened 2023-10-04
feat: Support for Robot Servers (Syself Fork)


apricote:
## Summary

We intend to merge the changes from [syself/hetzner-cloud-controller-manager](https://github.com/syself/hetzner-cloud-controller-manager) to provide support for Hetzner Robot (Dedicated/Bare Metal) servers in addition to cloud servers.

This issue will track the various tasks necessary.

## Subtasks

- [x] #525
- [x] #526
- [x] #527
- [x] #528

## Design Doc

We have written an internal Design Doc to figure out what exactly this means and how we want to tackle the different aspects. For transparency you can read it by expanding the section below.

<details>
<summary>Design Doc</summary>

#### Motivation

hcloud-cloud-controller-manager (HCCM) is used to expose certain Cloud functionality to Kubernetes Clusters. This includes Node Metadata, Network Routes & Load Balancers.

Right now, HCCM only supports the Hetzner Cloud API & cloud servers. Many customers have hybrid clusters running on Hetzner Dedicated (Robot) & Cloud servers. The underlying library of HCCM `kubernetes/cloud-provider` only works, if all nodes in the cluster can be managed by a single implementation. Effectively, this means, that clusters with a current version of hccm remove any non-Cloud Nodes from the Kubernetes Cluster.

The Hetzner Cloud API already works with Robot in a few scenarios:
- Robot servers can be added as Load Balancer Targets
- Robot servers can be attached to private Networks through vSwitches

In addition, we can use the Robot API to:
- Report Node Metadata

We have had a number of requests for this feature through [GitHub](https://github.com/hetznercloud/hcloud-cloud-controller-manager/issues/330).

There already exists a [fork of HCCM by Syself](https://github.com/syself/hetzner-cloud-controller-manager) that implements Robot support. They have offered to "donate" this to us, so we can use it as a starting point.

##### Robot Client

The Robot team provides a Rest API: https://robot.hetzner.com/doc/webservice/en.html#preface

The Robot team does not publish their own Go API Client. There is an [open source client](https://github.com/syself/hrobot-go) that is partially maintained by Syself. It is currently used in the Syself HCCM fork. As there is no better option, we will continue to use it.

We need to create and inject the client in our application similar to the hcloud-go client. This needs to be optional to not break existing setups.

##### Testing

We will need to set up a new CI workflow to verify the Robot support with a dedicated server. We might need to add some additional test cases to validate that it works across the different servers.

[...]

PR #532, opened 2023-10-06
refactor: duplicate utility to get server ID from ProviderID

apricote:
This utility function was duplicated with nearly the exact same functionality. This commit cleans it up by extracting to a new package (to avoid cyclic imports).

These two methods are about to get more complicated with #523, better to clean it up now than to make changes to both locations in the future.


internal/providerid/providerid.go
```diff
@@ -0,0 +1,36 @@
...
+const (
+	// providerPrefix is the prefix for all provider IDs. It MUST not be changed,
+	// otherwise existing nodes will not be recognized anymore.
+	providerPrefix = "hcloud://"
+)
+
+// ToServerID converts a ProviderID to a server ID.
+func ToServerID(providerID string) (int64, error) {
+	if !strings.HasPrefix(providerID, providerPrefix) {
+		return 0, fmt.Errorf("providerID does not have the expected prefix hcloud://: %s", providerID)
```
jooola, line 18:
```suggestion
		return 0, fmt.Errorf("providerID does not have the expected prefix %s: %s", providerPrefix, providerID)
```

jooola (commented):
This removes the`metrics.OperationCalled.WithLabelValues(op).Inc()` metrics, is that intended?

apricote:
> This removes the`metrics.OperationCalled.WithLabelValues(op).Inc()` metrics, is that intended?

This is intended. I do not think this metrics provides any value for this particular operation, as its just string parsing.

jooola (approved):

PR #537, opened 2023-10-17
feat(chart): add daemonset and node selector

simonostendorf:
This adds the ability to run hccm as a `DaemonSet` like other cloud controller managers. 
There are also nodeSelectors as a new value to set where the DaemonSet (and deployment) should run. 

The default deployment kind of `Deployment` with 1 replica is not touched. 

The new configuration options are documented in the README file.

### DIFF
--- a/chart/templates/daemonset.yaml
+++ b/chart/templates/daemonset.yaml
@@ -1,11 +1,10 @@
-{{- if eq $.Values.kind "Deployment" }}
+{{- if eq $.Values.kind "DaemonSet" }}
 apiVersion: apps/v1
-kind: Deployment
+kind: DaemonSet
 metadata:
   name: {{ include "hcloud-cloud-controller-manager.name" . }}
   namespace: {{ .Release.Namespace }}
 spec:
-  replicas: {{ .Values.replicaCount }}
   revisionHistoryLimit: 2
   selector:
     matchLabels:
--- a/chart/templates/daemonset.yaml
+++ b/chart/templates/daemonset.yaml
@@ -38,6 +38,11 @@ spec:
         {{- if gt (len .Values.additionalTolerations) 0 }}
         {{ toYaml .Values.additionalTolerations | nindent 8 }}
         {{- end }}
+      
+      {{- if gt (len .Values.nodeSelector) 0 }}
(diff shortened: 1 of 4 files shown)


chart/templates/daemonset.yaml
```diff
@@ -0,0 +1,88 @@
...
+            - "--{{ $key }}{{ if $value }}={{ $value }}{{ end }}"
+            {{- end }}
+            {{- end }}
+            {{- if $.Values.networking.enabled }}
+            - "--allocate-node-cidrs=true"
+            - "--cluster-cidr={{ $.Values.networking.clusterCIDR }}"
+            {{- end }}
+            {{- if (eq (int $.Values.replicaCount) 1) }}
+            - "--leader-elect=false"
+            {{- end }}
```
apricote, line 65: Running as a DaemonSet would definitely always require leader election. As the DaemonSet does not use the replicaCount, we should not use it to trigger flags.

```suggestion
```

apricote (commented):
Could you explain why its beneficial to run it as a DaemonSet vs Deployment? With the leader election, there is only one Pod doing the actual work.

chart/templates/daemonset.yaml
simonostendorf: Would it be better to trigger the flag if
a) kind is deployment and replicaCount is greater than 1
b) or if kind is daemonset?

This would prevent the user from forgetting to activate the leader election.

simonostendorf:
> Could you explain why its beneficial to run it as a DaemonSet vs Deployment? With the leader election, there is only one Pod doing the actual work.

Ich bin es gewöhnt (vorallem durch andere Cloud Controller Manager), das zentrale Komponenten als DaemonSet auf allen ControlPlane Nodes laufen. Einen wirklichen Mehrwert der Deploymentstrategie sehe ich nicht direkt.

chart/templates/daemonset.yaml
apricote: Leader Election is the default, right now we only deactivate it on Deployments if `replicaCount == 1`

apricote (approved):
Thank you very much @simonostendorf!

Issue #525, opened 2023-10-04
feat: Port & Review Changes from Syself fork


apricote:
Cherry-pick the relevant commits from the syself fork & make them work against current `main`.

Commits: https://github.com/hetznercloud/hcloud-cloud-controller-manager/compare/main...syself:hetzner-cloud-controller-manager:main

PR #546, opened 2023-10-25
refactor: move config parsing & validation to seperate package

apricote:
This significantly reduces the scope of the `hcloud.newCloud()` method, and allows us to cleanly introduce new validations for the Robot support.

There are some further refactorings that this change enables, but I want to publish them as a separate PR to make reviewing easier.

Related to #525

feat(config): stricter validation for settings `HCLOUD_LOAD_BALANCERS_ENABLED`, `HCLOUD_METRICS_ENABLED` & `HCLOUD_NETWORK_ROUTES_ENABLED`


internal/config/config.go
```diff
@@ -0,0 +1,177 @@
...
+	Address string
+}
+
+type AddressFamily string
+
+const (
+	AddressFamilyDualStack AddressFamily = "dualstack"
+	AddressFamilyIPv6      AddressFamily = "ipv6"
+	AddressFamilyIPv4      AddressFamily = "ipv4"
+)
```
jooola, line 50: This was moved in the config package, but I don't see the deletion from the previous location.

hcloud/cloud.go
```diff
@@ -61,107 +44,96 @@ type cloud struct {
...
 }
 
 func newCloud(_ io.Reader) (cloudprovider.Interface, error) {
 	const op = "hcloud/newCloud"
 	metrics.OperationCalled.WithLabelValues(op).Inc()
 
-	token := os.Getenv(hcloudTokenENVVar)
-	if token == "" {
-		return nil, fmt.Errorf("environment variable %q is required", hcloudTokenENVVar)
+	configuration, err := config.Read()
```
jooola, line 55: I always prefer to have explicit variable names, but for this one, I would propose to use `cfg` as standard.

internal/config/config.go
```diff
@@ -0,0 +1,177 @@
...
+	var config HCCMConfiguration
+
+	config.HCloudClient.Token = os.Getenv(hcloudToken)
+	config.HCloudClient.Endpoint = os.Getenv(hcloudEndpoint)
+	config.HCloudClient.Debug, err = getEnvBool(hcloudDebug)
+	if err != nil {
+		errs = append(errs, err)
+	}
+
+	config.Metrics.Enabled = os.Getenv(hcloudMetricsEnabled) != "false"
```
jooola, line 98: Why not use `getEnvBool()` ?

internal/config/config.go
```diff
@@ -0,0 +1,177 @@
...
+	config.Metrics.Address = hcloudMetricsAddress
+
+	// Validation happens in [HCCMConfiguration.Validate()]
+	config.Instance.AddressFamily = AddressFamily(os.Getenv(hcloudInstancesAddressFamily))
+	if config.Instance.AddressFamily == "" {
+		config.Instance.AddressFamily = AddressFamilyIPv4
+	}
+
+	config.LoadBalancer.Enabled = os.Getenv(hcloudLoadBalancersEnabled) != "false"
```
jooola, line 107: Why not use `getEnvBool()` ?

internal/config/config.go
```diff
@@ -0,0 +1,177 @@
...
+		errs = append(errs, err)
+	}
+
+	config.Network.NameOrID = os.Getenv(hcloudNetwork)
+	config.Network.DisableAttachedCheck, err = getEnvBool(hcloudNetworkDisableAttachedCheck)
+	if err != nil {
+		errs = append(errs, err)
+	}
+
+	config.Route.Enabled = os.Getenv(hcloudNetworkRoutesEnabled) != "false"
```
jooola, line 129: Why not use `getEnvBool()` ?

internal/config/config.go
apricote: This (and the other two codepaths you commented) was done for two reasons:
- **Backwards compatibility**: We previously interpreted everything but `"false"` as `true`, `getEnvBool()` has other rules
- **Default**: `getEnvBool()` defaults to `false`, while we want these fields to default to `true`.

I will add a comment to the function to explain this.

hcloud/cloud.go
```diff
@@ -61,107 +44,96 @@ type cloud struct {
 	client       *hcloud.Client
 	instances    *instances
 	loadBalancer *loadBalancers
+	config       config.HCCMConfiguration
```
jooola, line 47: Does it make sens to use a global here ?

apricote: Perhaps, I will do some more refactoring on the usage of `HCCMConfiguration` in `hcops` and the other packages in a follow up PR and will reevaluate then.

internal/config/config.go
apricote: Updated to your suggestion, will also add a note in the release notes for this.

jooola (approved):

PR #560, opened 2023-11-13
fix(deployment): remove dns-policy

### LINKED ISSUE #559: DnsPolicy results in crashing istio proxy
I am using the hcloud cloud controller manager together with the istio sidecar proxy. 

The istio proxy is stuck in a crash loop because it can't find the istiod host (`Error while dialing: dial tcp: lookup istiod.istio-system.svc on 185.12.64.1:53: no such host`). This is using the hetzner dns here (where obviously the entry can not be found).

If I switch the dnsPolicy inside the deployment to `ClusterFirst` everything works as expected. 

Is there a reason why it is configured to use `Default`?

See [here](https://github.com/hetznercloud/hcloud-cloud-controller-manager/blob/main/chart/templates/deployment.yaml#L19), [here](https://github.com/hetznercloud/hcloud-cloud-controller-manager/blob/main/deploy/ccm.yaml#L41), [here](https://github.com/hetznercloud/hcloud-cloud-controller-manager/blob/main/deploy/ccm-networks.yaml#L41) and [here](https://github.com/hetznercloud/hcloud-cloud-controller-manager/blob/main/chart/templates/daemonset.yaml#L18).

simonostendorf:
Closes #559

### DIFF
--- a/chart/templates/daemonset.yaml
+++ b/chart/templates/daemonset.yaml
@@ -15,7 +15,6 @@ spec:
         {{- include "hcloud-cloud-controller-manager.selectorLabels" . | nindent 8 }}
     spec:
       serviceAccountName: {{ include "hcloud-cloud-controller-manager.name" . }}
-      dnsPolicy: Default
       tolerations:
         # Allow HCCM itself to schedule on nodes that have not yet been initialized by HCCM.
         - key: "node.cloudprovider.kubernetes.io/uninitialized"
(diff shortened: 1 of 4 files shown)


apricote (approved):
Ran the e2e tests locally, all looking good :) Thanks for the fix @simonostendorf!

apricote:
This has caused a regression in our dev/test system for Deployments not using the Network Routes (#561):

1. Cluster is initialized
2. CoreDNS is not starting because the Node is still tainted with `node.cloudprovider.kubernetes.io/uninitialized`
3. HCCM is deployed, but can not initialize Node because it needs to resolve the domain `api.hetzner.cloud` with CoreDNS. This fails with `Get "https://api.hetzner.cloud/v1/servers?": dial tcp: lookup api.hetzner.cloud: i/o timeout`
4. => Deadlock

I am going to revert this for now to (finally) get #561 in, then we can take another look at how we can improve the situation for your sidecars.

simonostendorf:
> I am going to revert this for now to (finally) get #561 in, then we can take another look at how we can improve the situation for your sidecars.

Are there any news?

Issue #527, opened 2023-10-04
feat(helm): Add Robot Options to Helm Chart


apricote:
Support for Robot Servers will come with a number of new options to enable/disable specific features. These need to be exposed in the Helm Chart.

apricote:
Implemented in #561.

Issue #526, opened 2023-10-04
ci: Setup CI for Robot Servers


apricote:
Tasks:

- Setup locking of dedicated servers for CI
  - Using GitHub Actions [`concurrency`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency)
- Configure additional pipelines that also add the dedicated server to the Kubernetes Cluster
  - Should reinstall servers using [`installimage`](https://docs.hetzner.com/robot/dedicated-server/operating-systems/installimage) with [`autosetup`](https://docs.hetzner.com/robot/dedicated-server/operating-systems/installimage#autosetup). The node is then joined to our existing cluster using `k3sup`
- Write some tests for the new scenarios

Tests should include:
- Node does not get removed from cluster
- Load Balancer Target
- Node Metadata

apricote:
Implemented in #561.

PR #561, opened 2023-11-13
feat: support for Robot servers

### LINKED ISSUE #525: feat: Port & Review Changes from Syself fork

apricote:
Based on the [Fork by Syself](https://github.com/syself/hetzner-cloud-controller-manager) and the Design Doc in https://github.com/hetznercloud/hcloud-cloud-controller-manager/issues/523#issue-1925577588.

This ports most of the features the Fork while refactoring them to match our coding style and the improvements I made in preparation for this.

I recommend reviewing by commit.

Docs are still being worked on and will be added in a follow up PR before making a release.

Closes #525 #526 #527


hcloud/instances.go
```diff
@@ -20,48 +20,99 @@ import (
...
+		return nil, fmt.Errorf("found both a cloud & robot server for name %q", node.Name)
+	case cloudServer != nil && rServer == nil:
+		return hcloudServer{cloudServer}, nil
+	case cloudServer == nil && rServer != nil:
+		return robotServer{rServer, i.robotClient}, nil
+	default:
+		// Both nil
+		return nil, nil
+	}
 }
```
apricote, line 116: I dont love how this turned out. Open to any suggestions on improving it (and any sub functions)

chart/values.yaml
```diff
@@ -30,6 +30,19 @@ env:
         name: hcloud
         key: token
 
+  ROBOT_USER_NAME:
```
jooola, line 33: Since we use the `h` `cloud` convention to refer to Hetzner Cloud resources, what about using `h` `robot` to refer to any Hetzner Robot resources? 

e.g HCLOUD_TOKEN => HROBOT_USER_NAME

Or we should have CLOUD_TOKEN => ROBOT_USER_NAME.

hack/robot-e2e/ansible.cfg
```diff
@@ -0,0 +1,7 @@
+[defaults]
+inventory = ${PWD}/inventory.yml
+forks = 10
+host_key_checking = False
+
+[ssh_connection]
+# pipelining = True
```
jooola, line 7:
```suggestion
pipelining = True
```

This can speed up the ansible part. Unless it fails hard, I would enable it. Or completely remove it.

hack/robot-e2e/ansible.cfg
```diff
@@ -0,0 +1,7 @@
+[defaults]
+inventory = ${PWD}/inventory.yml
+forks = 10
```
jooola, line 3: I think we can use the default and remove this. We don't have many hosts to manage.

hack/robot-e2e/e2e-setup-robot-server.yml
```diff
@@ -0,0 +1,96 @@
+---
+- name: Prepare Reinstall
+  hosts: localhost
+  connection: local
+  gather_facts: false
+
+  vars:
+    scope: dev
```
jooola, line 8: What is this scope used for ?

hack/robot-e2e/e2e-setup-robot-server.yml
```diff
@@ -0,0 +1,96 @@
...
+      hetzner_password: "{{ lookup('ansible.builtin.env', 'ROBOT_PASSWORD') }}"
+
+  tasks:
+    - name: Get Server Info
+      community.hrobot.server_info:
+        server_number: "{{ server_number }}"
+      register: server_info
+
+    - name: Set Server Facts
+      ansible.builtin.set_fact: server_ip="{{ server_info.servers[0].server_ip }}" server_name="{{ server_info.servers[0].server_name }}"
```
jooola, line 24:
```suggestion
      ansible.builtin.set_fact:
        server_ip: "{{ server_info.servers[0].server_ip }}"
        server_name: "{{ server_info.servers[0].server_name }}"
```

hcloud/instances.go
```diff
@@ -20,48 +20,99 @@ import (
...
+	}
+
+	// If the node has no provider ID we try to find the server by name from
+	// both sources. In case we find two servers, we return an error.
+	cloudServer, err := getCloudServerByName(ctx, i.client, node.Name)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get hcloud server %q: %w", node.Name, err)
+	}
+
+	var rServer *robotmodels.Server
```
jooola, line 97:
```suggestion
	var robotServer *robotmodels.Server
```

hcloud/instances.go
jooola: You could extract the first error check in a separate if, and use the switch case to only check if a var is set.

hcloud/instances.go
```diff
@@ -20,48 +20,99 @@ import (
...
+	case cloudServer != nil && rServer != nil:
+		return nil, fmt.Errorf("found both a cloud & robot server for name %q", node.Name)
+	case cloudServer != nil && rServer == nil:
+		return hcloudServer{cloudServer}, nil
+	case cloudServer == nil && rServer != nil:
+		return robotServer{rServer, i.robotClient}, nil
+	default:
+		// Both nil
+		return nil, nil
+	}
```
jooola, line 115:
```suggestion

	if cloudServer != nil && rServer != nil {
		return nil, fmt.Errorf("found both a cloud & robot server for name %q", node.Name)
	}
	switch {
	case cloudServer != nil:
		return hcloudServer{cloudServer}, nil
	case rServer != nil:
		return robotServer{rServer, i.robotClient}, nil
	default:
		return nil, nil
	}
```

hcloud/instances.go
```diff
@@ -154,3 +215,80 @@ func nodeAddresses(addressFamily config.AddressFamily, networkID int64, server *
...
+type genericServer interface {
+	Shutdown() (bool, error)
+	Metadata(addressFamily config.AddressFamily, networkID int64) (*cloudprovider.InstanceMetadata, error)
+}
+
+type hcloudServer struct {
+	*hcloud.Server
+}
+
+func (s hcloudServer) Shutdown() (bool, error) {
```
jooola, line 256:
```suggestion
func (s hcloudServer) IsShutdown() (bool, error) {
```

hcloud/instances.go
```diff
@@ -154,3 +215,80 @@ func nodeAddresses(addressFamily config.AddressFamily, networkID int64, server *
...
+		Region:        s.Datacenter.Location.Name,
+	}, nil
+}
+
+type robotServer struct {
+	*robotmodels.Server
+	robotClient robot.Client
+}
+
+func (s robotServer) Shutdown() (bool, error) {
```
jooola, line 275:
```suggestion
func (s robotServer) IsShutdown() (bool, error) {
```

hcloud/util.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+	return server, nil
+}
+
+func getRobotServerByName(c robot.Client, node *corev1.Node) (server *robotmodels.Server, err error) {
+	const op = "hcloud/getRobotServerByName"
+
+	if c == nil {
+		return nil, errMissingRobotCredentials
+	}
```
jooola, line 60: This might mislead us in the future, if we do not configure the robot client for some other reason.

Could we have an error such as "the robot client is not configured" instead ?

hcloud/util.go
```diff
@@ -0,0 +1,129 @@
...
+
+	// check for rate limit
+	if robot.IsRateLimitExceeded() {
+		return nil, fmt.Errorf("%s: rate limit exceeded - next try at %q", op, robot.TimeOfNextPossibleAPICall().String())
+	}
```
jooola, line 65: What is the recommended behavior ? Should we fail early or block until the limit is down ? I assume this is part of a reconciliation loop, so maybe failing early is best ? What happens if we stop in the middle of the reconciliation and the state is only half applied ?

internal/config/config.go
```diff
@@ -36,6 +42,14 @@ type HCloudClientConfiguration struct {
 	Debug    bool
 }
 
+type RobotClientConfiguration struct {
```
jooola, line 45:
```suggestion
type RobotConfiguration struct {
```

Or move the enabled flag out of the struct ?

internal/config/config.go
```diff
@@ -97,6 +112,22 @@ func Read() (HCCMConfiguration, error) {
 		errs = append(errs, err)
 	}
 
+	cfg.Robot.Username = os.Getenv(robotUserName)
+	cfg.Robot.Password = os.Getenv(robotPassword)
+	// Enable Robot Support when both credentials are set
+	cfg.Robot.Enabled = cfg.Robot.Username != "" && cfg.Robot.Password != ""
```
jooola, line 118: Is it not better to have a flag to enable robot, and fail hard on missing credentials ?

internal/providerid/providerid.go
```diff
@@ -7,30 +7,66 @@ import (
...
+//
+// This method supports all formats for the ProviderID that were ever used.
+// If a format is ever dropped from this method the Nodes that still use that
+// format will get abandoned and can no longer be processed by HCCM.
+func ToServerID(providerID string) (id int64, isCloudServer bool, err error) {
+	if !strings.HasPrefix(providerID, prefixCloud) &&
+		!strings.HasPrefix(providerID, prefixRobot) &&
+		!strings.HasPrefix(providerID, prefixRobotLegacy) {
+		return 0, false, fmt.Errorf("providerID does not have one of the the expected prefixes (%s, %s, %s): %s", prefixCloud, prefixRobot, prefixRobotLegacy, providerID)
+	}
```
jooola, line 37: What about moving this check in the default switch case below ?

internal/robot/client.go
```diff
@@ -0,0 +1,94 @@
...
+
+type cacheRobotClient struct {
+	robotClient hrobot.RobotClient
+	timeout     time.Duration
+
+	lastUpdate time.Time
+
+	// cache
+	l []robotmodels.Server
+	m map[int]*robotmodels.Server
```
jooola, line 18:
```suggestion
	serverList []robotmodels.Server
	serverMap map[int]*robotmodels.Server
```

internal/robot/client.go
```diff
@@ -0,0 +1,94 @@
...
+
+		// remove all entries from map and populate it freshly
+		c.m = make(map[int]*robotmodels.Server)
+		for i, server := range list {
+			c.m[server.ServerNumber] = &list[i]
+		}
+
+		// set time of last update
+		c.lastUpdate = time.Now()
+	}
```
jooola, line 46: I think this can be deduplicated, the same is in the next function. Maybe move it into shouldSync and rename it  RefreshCache ?

internal/robot/ratelimit.go
```diff
@@ -0,0 +1,69 @@
...
+	"time"
+
+	robotmodels "github.com/syself/hrobot-go/models"
+)
+
+func InitRateLimit(rateLimitWaitTime time.Duration) {
+	rateLimit = rateLimitHandler{
+		waitTime: rateLimitWaitTime,
+	}
+}
```
jooola, line 14: What about using https://pkg.go.dev/k8s.io/client-go@v0.28.3/util/flowcontrol#Backoff ?

jooola (commented):
Made a first pass, I think some small things could be made more consistent.

hack/robot-e2e/e2e-setup-robot-server.yml
apricote: The scope is introduced in `hack/dev-up.sh` to deploy multiple clusters into the same project (ie name prefix). This was more relevant before we had unique projects per CI pipeline.

hcloud/instances.go
apricote: The identifier `robotServer` is already used by the struct implementing `genericServer`.

internal/robot/client.go
apricote: Called it `servers` & `serversByID`

internal/robot/client.go
apricote: Refactored it a bit :)

internal/robot/ratelimit.go
apricote: This is using an exponential backoff instead of the constant backoff we have right now. I am tempted to just remove the whole Robot ratelimiting section for now and push it to a follow up PR (where we might also rething Hetzner Cloud rate limiting, which we do not handle at all right now).

apricote: Removed it from this Patch series, will submit a follow up PR to add it back in in a less invasive manner using standard tooling.

hcloud/util.go
apricote: Yes, all of this is called in a reconciliation loop so failing early is preferred.

If we stop in the middle, only the state until then is applied and everything else will get applied with the next loop. This is not really a problem here, because the reconciliation loop is ran for each node separately.

internal/config/config.go
apricote: Nice idea, implemented in next push :)

hcloud/util.go
apricote: New error: `no robot client configured, make sure to enable Robot support in the configuration`

chart/values.yaml
apricote: In my mind, `hcloud` is an established abbreviation for Hetzner Cloud, as used in the `cli` and this projects name and many other places.

I have never heard anyone call it `hrobot` (besides the Go SDK). Its always Robot or Hetzner Robot.

This is also a compatibility thing, these [env var names were used in the Syself fork](https://github.com/syself/hetzner-cloud-controller-manager/blob/04b701babcebdffda9cb869e24416cc9f8998fc3/hcloud/cloud.go#L53-L54), so I would prefer to keep them as they are.

apricote:
Implemented all review feedback. Diff here: https://github.com/hetznercloud/hcloud-cloud-controller-manager/compare/dee5e961e27d41f309292c45b3c011ef107c8a4b..2af7752e1d6a85a4062584d5e57aedc70cad4f7b

hcloud/instances.go
jooola: Maybe use the `h` to determine if it's a local struct or a struct from the hrobot library ?

apricote:
Tests failed before as I forgot to add the `ROBOT_ENABLED` flag to the helm chart: https://github.com/hetznercloud/hcloud-cloud-controller-manager/compare/2af7752e1d6a85a4062584d5e57aedc70cad4f7b..32491034b7545c00a95ede32b3a9e80a4376edb8

internal/robot/client.go
```diff
@@ -0,0 +1,92 @@
...
+		return c.updateCache()
+	}
+
+	return nil
+}
+
+func (c *cacheRobotClient) shouldSync() bool {
+	// map is nil means we have no cached value yet
+	if c.serversByID == nil {
+		c.serversByID = make(map[int]*hrobotmodels.Server)
```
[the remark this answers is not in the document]
apricote: This only happens once when the struct is freshly created. I have just moved it into `NewClient()`.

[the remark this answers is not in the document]
apricote: Added a mutex for this.

chart/values.yaml
```diff
@@ -30,6 +30,19 @@ env:
...
+    valueFrom:
+      secretKeyRef:
+        name: hcloud
+        key: robot-user
```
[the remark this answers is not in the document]
apricote: Updated all to to match the secret key :)

hcloud/instances.go
```diff
@@ -154,3 +221,80 @@ func nodeAddresses(addressFamily config.AddressFamily, networkID int64, server *
 	}
 	return addresses
 }
+
+func robotNodeAddresses(addressFamily config.AddressFamily, server *hrobotmodels.Server) []corev1.NodeAddress {
```
jooola, line 225: Can both `hcloudNodeAddresses` and `robotNodeAddresses` be part of the genericServer Interface ? If so, let's postpone that for a future PR.

hack/dev-up.sh
```diff
@@ -164,16 +164,28 @@ if [[ -n "${DEBUG:-}" ]]; then set -x; fi
     # Create HCLOUD_TOKEN Secret for hcloud-cloud-controller-manager.
     ( trap error ERR
       if ! kubectl -n kube-system get secret hcloud >/dev/null 2>&1; then
-        kubectl -n kube-system create secret generic hcloud --from-literal="token=$HCLOUD_TOKEN" --from-literal="network=$scope_name"
+        data=(
+          --from-literal="token=$HCLOUD_TOKEN"
+          --from-literal="network=$scope_name"
+        )
+        if [[ -v ROBOT_USER ]]; then
```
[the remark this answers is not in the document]
apricote: It now needs to be set anywhere, but this is kind of unrelated to whether Robot Support is enabled or not. You might still decide to enable it at a later point through the Helm Chart, but the secret is only setup once in `dev-up.sh`.

As you can also see here, for the credentials I opted for a "if available, pass through, if not, just continue as normal" approach:

https://github.com/hetznercloud/hcloud-cloud-controller-manager/pull/561/files#diff-a9d07052701ab3b718c82ac0fe74b965a8260b06bb5906362081718b0ee593b4R35-R46

hcloud/instances.go
apricote: For sure, Ill put it on my list.

apricote:
Waiting on #564 for a final rebase & CI fixing run.

hack/robot-e2e/ansible.cfg
```diff
@@ -0,0 +1,6 @@
+[defaults]
+inventory = ${PWD}/inventory.yml
+host_key_checking = False
+
```
jooola, line 4:
```suggestion
[defaults]
inventory = ${PWD}/inventory.yml
host_key_checking = False
stdout_callback = community.general.yaml
```

This should improve the ansible logs, by printing yaml instead of json. You might also need to add the `community.general` collection to the requirements.yml file.

.github/workflows/test_e2e.yml
```diff
@@ -102,3 +103,123 @@ jobs:
...
+      - name: Setup Robot Server
+        env:
+          ROBOT_USER: ${{ secrets.ROBOT_USER }}
+          ROBOT_PASSWORD: ${{ secrets.ROBOT_PASSWORD }}
+        working-directory: hack/robot-e2e
+        run: |
+          ansible-galaxy install -r requirements.yml
+          echo "::group::ansible-playbook e2e-setup-robot-server.yml"
+          ansible-playbook e2e-setup-robot-server.yml -e scope=$SCOPE -e server_number=$SERVER_NUMBER -vvv
+          echo "::endgroup::"
```
jooola, line 177:
```suggestion
      - name: Setup Robot Server
        env:
          ROBOT_USER: ${{ secrets.ROBOT_USER }}
          ROBOT_PASSWORD: ${{ secrets.ROBOT_PASSWORD }}
          PY_COLORS: true
          ANSIBLE_FORCE_COLOR: true
        working-directory: hack/robot-e2e
        run: |
          ansible-galaxy install -r requirements.yml
          echo "::group::ansible-playbook e2e-setup-robot-server.yml"
          ansible-playbook e2e-setup-robot-server.yml -e scope=$SCOPE -e server_number=$SERVER_NUMBER -vvv
          echo "::endgroup::"
```

This should add some colors the ansible logs, and help to debug.

.github/workflows/test_e2e.yml
```diff
@@ -102,3 +103,123 @@ jobs:
...
+          skaffold build --tag="e2e-${GITHUB_RUN_ID}-${GITHUB_RUN_NUMBER}"
+          tag=$(skaffold build --tag="e2e-${GITHUB_RUN_ID}-${GITHUB_RUN_NUMBER}" --quiet --output="{{ (index .Builds 0).Tag }}")
+          skaffold deploy \
+            --profile=robot \
+            --images=hetznercloud/hcloud-cloud-controller-manager=$tag
+
+      - name: Setup Robot Server
+        env:
+          ROBOT_USER: ${{ secrets.ROBOT_USER }}
+          ROBOT_PASSWORD: ${{ secrets.ROBOT_PASSWORD }}
```
jooola, line 171:
```suggestion
        env:
          ROBOT_USER: ${{ secrets.ROBOT_USER }}
          ROBOT_PASSWORD: ${{ secrets.ROBOT_PASSWORD }}
          PY_COLORS: true
          ANSIBLE_FORCE_COLOR: true
```

apricote: Thanks :)

Issue #562, opened 2023-11-15
ci: improve observability into e2e tests


apricote:
Our e2e test suite right now has some issues which make debugging failed tests harder then necessary. There is also an annoying bug with the Actions UI where logs are sometimes not shown, so you don't even know where in the process you are. These improvements could be made:

## Split step `Run tests` into multiple.

The `Run tests` step currently sets up the test environment, runs the tests and tears it down again:

https://github.com/hetznercloud/hcloud-cloud-controller-manager/blob/2c8378647923d3758bef63bc96dd83c5cb422f85/.github/workflows/test_e2e.yml#L35-L52

This could be split into the following steps to get a better visibility into the process:

1. Setup dev env
2. Skaffold
3. Run tests
4. Tear down dev env

## Persist Logs

Right now, after the tests are done we tear down the cluster and loose access to logs from all Pods. To find out what went wrong, we need to start the tests locally and reproduce to get the logs.

It would be very helpful to download & save the _HCCM Pod Logs_ & _Events_ as artifacts before taking down the cluster.

## Deduplicate Code into repo actions

With #561 we are going to have lots of duplicate steps for the e2e tests between Cloud & Robot. We can refactor this to use Actions that are saved in this repository.

apricote:
Started implementation in #564

apricote:
Decided against refactoring the CI code into an action for the reasons explained here: https://github.com/hetznercloud/hcloud-cloud-controller-manager/pull/573#issuecomment-1838568864

The two other tasks are already implemented.

PR #564, opened 2023-11-17
ci(e2e): persist logs for improved observability (#562)

apricote:
### [ci(e2e): persist logs for improved observability](https://github.com/hetznercloud/hcloud-cloud-controller-manager/commit/425c448460dad46161e541a541131c6e34b6abc9)

Download and save relevant logs generated during the e2e tests to make debugging them easier. Right now we have no visibility into the cluster, which makes debugging impossible and we always need to replicate locally before getting any kind of useful insights.

Saved data includes:

- Logs of HCCM Pods
- Any kind of Events generated in Kubernetes

The data is made available through GitHub Actions artifacts.

### [ci(e2e): split jobs](https://github.com/hetznercloud/hcloud-cloud-controller-manager/commit/13add1fae9a0b2604b948d014dbe130329cc7b0a)

Split the large job into multiple smaller jobs doing distinctive things.

This helps with a bug in the GitHub Actions UI, where the logs sometimes do not load. With separate jobs you at least know where in the process you are.

### DIFF
--- a/.github/workflows/test_e2e.yml
+++ b/.github/workflows/test_e2e.yml
@@ -62,7 +62,43 @@ jobs:
         run: |
           go test ./tests/e2e -tags e2e -v -timeout 60m
 
+      - name: Download logs & events
+        if: always()
+        continue-on-error: true
+        run: |
+          mkdir debug-logs
+          kubectl logs \
+            --namespace kube-system \
+            --selector app.kubernetes.io/name=hcloud-cloud-controller-manager \
+            --all-containers \
+            --prefix=true \
+            --tail=-1 \
+            > debug-logs/hccm.log
+          
+          kubectl get events \
+            --all-namespaces \
+            --sort-by=.firstTimestamp \
+            --output yaml \
+          > debug-logs/events.yaml
+
+      - name: Show HCCM Logs on Failure
+        if: failure()
+        continue-on-error: true
+        run: |
+          echo "::group::hccm.log"
+          cat debug-logs/hccm.log
+          echo "::endgroup::"
+
       - name: Cleanup test environment
         if: always()
+        continue-on-error: true
         run: |
           hack/dev-down.sh
+
+      - name: Persist debug artifacts
+        if: always()
+        continue-on-error: true
+        uses: actions/upload-artifact@v3
+        with:
+          name: debug-logs-${{ env.SCOPE }}
+          path: debug-logs/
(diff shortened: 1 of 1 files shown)


.github/workflows/test_e2e.yml
```diff
@@ -31,22 +44,57 @@ jobs:
...
-
-      - name: Run tests
-        env:
-          K3S_CHANNEL: ${{ matrix.k3s }}
-          SCOPE: gha-${{ github.run_id }}-${{ github.run_attempt }}-${{ matrix.k3s }}
-
-          # Domain must be available in the account running the tests. This domain is available in the account
-          # running the public integration tests.
-          CERT_DOMAIN: hc-integrations-test.de
+      - name: install k3sup
```
jooola, line 47:
```suggestion
      - name: Install k3sup
```

jooola (approved):

.github/workflows/test_e2e.yml
```diff
@@ -31,22 +44,57 @@ jobs:
...
+        if: always()
+        continue-on-error: true
+        run: |
+          kubectl logs \
+            --namespace kube-system \
+            --selector app.kubernetes.io/name=hcloud-cloud-controller-manager \
+            --all-containers \
+            --prefix=true \
+            --tail=-1 \
+            > debug-logs/hccm.log
```
jooola, line 80: Should we print the logs in the CI in case of failure ? For small errors one might be able to spot the mistake without having to download the artifacts ?

apricote: Sounds good, do you think we should log everything or just the hccm logs?

jooola: I'd go with the most useful one, but not both. HCCM logs seem to be the most useful ones?

apricote: I just put in the HCCM & hcloud API logs for now. There are too many events to do this nicely.

apricote: I put them into [log groups](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#grouping-log-lines) that should be collapsed by default. The cloud api logs are only logged if the env setup failed.

apricote:
I have dropped the debug logs from the `hcloud` CLI, as they way that the CLI handles the debug file does not really work when multiple processes are running simultaneously.

PR #565, opened 2023-11-18
chore: use semver with helm chart

### LINKED ISSUE #529: [bug] helm chart version is not semver compliant
As per https://helm.sh/docs/topics/charts/#the-chartyaml-file, a helm chart version has to be a SemVer 2.0.0 version.

The [hcloud/hcloud-cloud-controller-manager](https://github.com/hetznercloud/hcloud-cloud-controller-manager/tree/main/chart) chart is not a valid helm chart currently since it does not use a SemVer 2.0.0 version (due to the `v` prefix).

kranurag7:
This commit updates the chart version of hcloud-cloud-controller-manager We only need to update the version in Chart.yaml because `goreleaser` already strips the `v` from {{ .Version }}

To patch the extra `v` with image tag, I updated the template to use `tag: 'v{{ $.Chart.Version }}'`

Ref: https://goreleaser.com/customization/release/#github 
Above, the name template looks something like this. 

`name_template: "{{.ProjectName}}-v{{.Version}} {{.Env.USER}}"`

So, IMHO, we don't need to make any change in `.goreleaser.yaml`
For testing and Ref: 
I created a snapshot release locally using `goreleaser release --snapshot` and the diff updates the chart version and image tag as per the requirement described in the issue. 

<details><summary> Attached Screenshot </summary>
<p>

image

</p>
</details> 

Fixes #529

### DIFF
--- a/chart/Chart.yaml
+++ b/chart/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v2
 name: hcloud-cloud-controller-manager
 type: application
-version: v1.18.0 # x-release-please-version
+version: 1.18.0 # x-release-please-version
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -32,7 +32,7 @@ env:
 
 image:
   repository: hetznercloud/hcloud-cloud-controller-manager
-  tag: '{{ $.Chart.Version }}'
+  tag: 'v{{ $.Chart.Version }}'
 
 monitoring:
   # When enabled, the hccm Pod will serve metrics on port :8233


apricote (changes requested):
Hey @kranurag7,

thanks for the PR! I think we do need to make some changes in the goreleaser configuration, because we explicitly added the `v` back into the version when calling `scripts/release-generate-deployment-yamls.sh` and `./scripts/publish-helm-chart.sh`. See my comment https://github.com/hetznercloud/hcloud-cloud-controller-manager/issues/529#issuecomment-1754642063

The first script actually overwrites the version in `Chart.yaml` to be whatever the script was called with, currently including the `v`:

https://github.com/hetznercloud/hcloud-cloud-controller-manager/blob/07a3df9cd16e4438d8448b6866edef7554ff6f15/scripts/release-generate-deployment-yamls.sh#L15

I also think this is worth pointing out in the release notes, so I will update the title of this PR to `feat(helm): remove "v" prefix from chart version`.

apricote:
Hey @kranurag7, do you have some time to work on this within the next week? Otherwise I would implement the fixes to make sure this will be included in the next version.

kranurag7:
This PR was always in back of my head but I was little busy. I'll for sure push the changes tomorrow.

apricote:
Yea, running the `goreleaser` config outside of CI (and without actually publishing something) is a PITA :(

Changes look good, i will push one more cleanup commit and then do a release-candidate to verify its working :rocket: 

Thanks for the patches!

apricote (approved):

PR #566, opened 2023-11-21
fix(deployment): revert #560: remove dns-policy

apricote:
This reverts commit 2c8378647923d3758bef63bc96dd83c5cb422f85 / #560.

The commit caused a regression where clusters not using the host network were unable to bootstrap successfully:

- CoreDNS did not start because the Node was still tainted as uninitalized
- HCCM could not initialize the Node because it failed to resolve the DNS entry for "api.hetzner.cloud".


apricote:
I am going to remove the changelog entries for both PRs, to reduce noise in the changelog.

jooola (approved):

Issue #569, opened 2023-11-29
feat: Emit Kubernetes Events for issues


apricote:
For most of our internal errors we just log them, but users rarely read the logs of HCCM. It would be nicer if we emitted Kubernetes Events attached to the resources where the error occured.

This was already being done in the syself fork, but removed for the merge to keep the initial scope small and reviewable.

apricote:
Was implemented in #598

PR #570, opened 2023-11-29
feat(load-balancer): handle planned targets exceedings max targets

apricote:
Catches the error condition before we even make a request to the API. This causes less requests to be sent to the Hetzner Cloud API.

This does not change the behavior, as we just log & continue right now if we encounter a `ErrorCodeResourceLimitExceeded` while adding the target. Both cases are just visible in the HCCM logs (see #569 for improvements in that regard).

### DIFF
--- a/internal/hcops/load_balancer.go
+++ b/internal/hcops/load_balancer.go
@@ -627,6 +627,8 @@ func (l *LoadBalancerOps) ReconcileHCLBTargets(
 		}
 	}
 
+	numberOfTargets := len(lb.Targets)
+
 	// Extract IDs of the hc Load Balancer's server targets. Along the way,
 	// Remove all server targets from the HC Load Balancer which are currently
 	// not assigned as nodes to the K8S Load Balancer.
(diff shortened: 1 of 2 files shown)


internal/hcops/load_balancer.go
```diff
@@ -731,6 +741,11 @@ func (l *LoadBalancerOps) ReconcileHCLBTargets(
 				continue
 			}
 
+			if numberOfTargets >= lb.LoadBalancerType.MaxTargets {
+				klog.InfoS("cannot add ip target because max number of targets have been reached", "op", op, "service", svc.ObjectMeta.Name, "targetName", k8sNodeNames[int64(id)])
```
jooola, line 745: Shouldn't this be a warning?

apricote: `klog` does not have an interface for structured logging with the warning level, only for `InfoS` and `ErrorS`. We can make this error if you prefer that.

Currently this triggers the `ErrorCodeResourceLimitExceeded` check a few lines down, where we also log to `InfoS`.

apricote:
Quick fix to the original issue description, the current behaviour is exactly the same, but we send less API requests in this case.

If an `ErrorCodeResourceLimitExceeded` error is encountered while adding the targets, we just continue with our life.

jooola (approved):

internal/hcops/load_balancer.go
jooola: If the desired state if not reached, I assume something is wrong, so I'd go with an error. But feel free to keep it as is.

PR #572, opened 2023-11-30
feat(robot): handle ratelimiting with constant backoff

apricote:
Add a constant backoff in case we are being rate limited by the Robot API.

This is done through an opaque wrapping of the `robot.Client` interface, so callers do not need to do this manually.


internal/robot/cache.go
```diff
@@ -21,7 +20,7 @@ type cacheRobotClient struct {
 	serversByID map[int]*hrobotmodels.Server
 }
 
-func NewClient(robotClient hrobot.RobotClient, cacheTimeout time.Duration) Client {
+func WithCache(cacheTimeout time.Duration, robotClient Client) Client {
```
jooola, line 23:
```suggestion
func NewCachedClient(robotClient Client, cacheTimeout time.Duration) Client {
```

internal/robot/ratelimit.go
```diff
@@ -0,0 +1,93 @@
...
+
+type rateLimitClient struct {
+	robotClient Client
+
+	waitTime    time.Duration
+	exceeded    bool
+	lastChecked time.Time
+}
+
+func WithRateLimit(rateLimitWaitTime time.Duration, robotClient Client) Client {
```
jooola, line 19:
```suggestion
func NewRateLimitedClient(robotClient Client, rateLimitWaitTime time.Duration) Client {
```

jooola (commented):
What about using https://pkg.go.dev/github.com/syself/hrobot-go#NewBasicAuthClientWithCustomHttpClient and implement the rate limiter and cache around the http.Client ?

internal/robot/cache.go
apricote: I put the client last on purpose. I found it pretty confusing to have the client first when nesting, as the argument `RateLimitWaitTime` is not next to `WithRateLimit` anymore. For comparison:

Client last (current version):

```go
robotClient = robot.WithRateLimit(
  robot.WithCache(
    cfg.Robot.CacheTimeout,
    c
  ),
  cfg.Robot.RateLimitWaitTime,
)
```

Client first (your suggestion):

```go
robotClient = robot.WithRateLimit(
  cfg.Robot.RateLimitWaitTime,
  robot.WithCache(
    cfg.Robot.CacheTimeout,
    c
  ),
)
```

apricote:
> What about using https://pkg.go.dev/github.com/syself/hrobot-go#NewBasicAuthClientWithCustomHttpClient and implement the rate limiter and cache around the http.Client ?

IMO both additional features are based on the business logic of the API, rather than the transport mechanism.

For caching, we actually only use the list endpoint, and serve `GET /servers/:id` from the list cache. We do not cache the `GET /reset/:id` call.

For Rate Limiting we would need to parse the response in our middleware once more to evaluate the error, and similarly fake out the http response if we decide not to make a proper request.

Implement this on top of the `Client` interface is also simpler than doing it on the more generalized [`http.RoundTripper`](https://pkg.go.dev/net/http#RoundTripper).

internal/robot/cache.go
apricote: Yea, I updated the naming :) Just did not agree with the changed order of arguments in your suggested replacement. (Sorry, should have been more explicit.)

PR #573, opened 2023-11-30
ci: deduplicate e2e code

### LINKED ISSUE #562: ci: improve observability into e2e tests

apricote:
The e2e jobs for cloud & robot shared most of their actual steps. This deduplicates the lines into a shared action.

Closes #562


apricote:
Closing this. If we migrate to a single composite action, we loose all the visibility into the process we previously gained by splitting the steps into multiple. Composite Actions are always rendered as a single step.

We could move to multiple actions, but then the code would be distributed over 3-4 files, making it even harder to understand whats happening.

PR #577, opened 2023-12-04
docs: describe networks ip ranges with example

### LINKED ISSUE #576: Clarification about network overlapping
Hello, reading through docs, especially this https://github.com/hetznercloud/hcloud-cloud-controller-manager/blob/main/docs/deploy_with_networks.md file, i got confused about following parts:

In example, it recommends creating network 10.0.0.0/8 (pretty wide one)
In same example, it says that cluster-cidr is using 10.244.0.0/16 (which falls under 10.0.0.0/8)

And in the end of doc, it states:

```
Solution: Make sure the cluster-cidr does not overlap with the Hetzner Cloud Network. For example your Subnetwork could be 10.10.10.0/24 when the cluster-cidr is set to 10.244.0.0/16.
```

So should it overlap with some of k8s networks or not? 

Say i created 10.121.0.0/16 network in hetzner, assigned 10.121.0.{2..9} IPs to node as internal IP, and decided my cluster cidr will be 10.251.0.0/16. Is that correct setup?

apricote:
Fixes #576

### DIFF
--- a/docs/deploy_with_networks.md
+++ b/docs/deploy_with_networks.md
@@ -35,6 +35,11 @@ Some example values:
     - Node 1: `10.0.16.0/24`, Node 2: `10.0.17.0/24`, ...
 - Service CIDR: `10.0.8.0/21` (up to 2046 `ClusterIP` services)
 
+Some further restrictions and considerations are explained in the Hetzner Cloud Docs:
+
+- [Which IP addresses can I use?](https://docs.hetzner.com/cloud/networks/faq#which-ip-addresses-can-i-use)
+- [Are any IP addresses reserved?](https://docs.hetzner.com/cloud/networks/faq/#are-any-ip-addresses-reserved)
+
 ## Common Issues
 
 ### FailedToCreateRoute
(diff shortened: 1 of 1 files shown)


docs/deploy_with_networks.md
```diff
@@ -19,12 +19,31 @@ When deploying Cilium, make sure that you have set `tunnel: disabled` and `nativ
...
 Error Message:
+
 ```
 Could not create route xy-xy-xy-xy-xy 10.244.0.0/24 for node xy.example.com after 1s: hcloud/CreateRoute: network route destination overlaps with another subnetwork or network route (invalid_input)
 ```
+
 Solution:
-Make sure the cluster-cidr does not overlap with the Hetzner Cloud Network.
-For example your Subnetwork could be `10.10.10.0/24` when the *cluster-cidr* is set to `10.244.0.0/16`.
+Make sure the cluster-cidr does not overlap with the Hetzner Cloud Subnet. Check [Considerations on the IP Ranges](#considerations-on-the-ip-ranges) for more information.
```
rlex, line 49: Maybe something like "Make sure the cluster-cidr does not overlap with any subnet created in Hetzner Cloud Network"?

pauhull (approved):

PR #590, opened 2023-12-15
feat: add ability to use node ip as LB target

blitss:
(sorry for any possible mistakes in this PR, I'm not really familiar with Go)
In this PR I added the ability to use node ip as the target for the Hetzner LB. The motivation behind this is that current Robot implementation only adds public IPs from the Robot to the LB, and someone might have cluster setup like this, which uses internal IP from the vswitch.

```
k get node -o wide
NAME    STATUS   ROLES           AGE     VERSION   INTERNAL-IP   EXTERNAL-IP   OS-IMAGE                KERNEL-VERSION   CONTAINER-RUNTIME
node1   Ready    control-plane   5h51m   v1.28.3   10.0.1.3      <none>        Talos (v1.6.0-beta.1)   6.1.65-talos     containerd://1.7.10
```

which just doesn't work with following implementation. I also added warning for when the name in the Robot and k8s don't match. It's opt-in and you have to specify `HCLOUD_LOAD_BALANCERS_USE_NODE_IP` to use that.

I ran the controller like 

```bash
 ROBOT_ENABLED=true \
HCLOUD_TOKEN= ROBOT_PASSWORD=ROBOT_USER="" \
KUBECONFIG=$HOME/.kube-old/config  \
HCLOUD_NETWORK=1047143 \
HCLOUD_NETWORK_ROUTES_ENABLED=false \
HCLOUD_NETWORK_DISABLE_ATTACHED_CHECK=true \
HCLOUD_LOAD_BALANCERS_USE_NODE_IP=true \
go run . --allow-untagged-cloud  --cloud-provider=hcloud --route-reconciliation-period=30s --webhook-secure-port=0 --leader-elect=false --kubeconfig=$HOME/.kube-old/config

```
and it worked like a charm for me. Result:

Pasted_Image_16_12_23__02_45

Since you're specifying a network I have to also use flags like `HCLOUD_NETWORK_ROUTES_ENABLED` and `HCLOUD_NETWORK_DISABLE_ATTACHED_CHECK`.

Would love to add test for it, but I think it's going to add a lot of complexities testing this along with a vswitch.

### DIFF
--- a/hcloud/instances_util.go
+++ b/hcloud/instances_util.go
@@ -19,6 +19,7 @@ package hcloud
 import (
 	"context"
 	"fmt"
+	"k8s.io/klog/v2"
 	"strings"
 
 	hrobotmodels "github.com/syself/hrobot-go/models"
@@ -91,6 +92,7 @@ func getRobotServerByID(c robot.Client, id int, node *corev1.Node) (*hrobotmodel
 
 	// check whether name matches - otherwise this server does not belong to the respective node anymore
 	if server.Name != node.Name {
+		klog.Warningf("%s: server %d has name %q, but node %q has name %q. if you want node to be matched with node in Hetzner Robot you should rename it.", op, id, server.Name, node.Name, node.Name)
 		return nil, nil
 	}
 
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -31,6 +31,7 @@ const (
 	hcloudLoadBalancersNetworkZone           = "HCLOUD_LOAD_BALANCERS_NETWORK_ZONE"
 	hcloudLoadBalancersDisablePrivateIngress = "HCLOUD_LOAD_BALANCERS_DISABLE_PRIVATE_INGRESS"
 	hcloudLoadBalancersUsePrivateIP          = "HCLOUD_LOAD_BALANCERS_USE_PRIVATE_IP"
+	hcloudLoadBalancersUseNodeIP             = "HCLOUD_LOAD_BALANCERS_USE_NODE_IP"
 	hcloudLoadBalancersDisableIPv6           = "HCLOUD_LOAD_BALANCERS_DISABLE_IPV6"
 
 	hcloudMetricsEnabled = "HCLOUD_METRICS_ENABLED"
@@ -74,6 +75,7 @@ type LoadBalancerConfiguration struct {
 	NetworkZone           string
 	DisablePrivateIngress bool
 	UsePrivateIP          bool
+	UseNodeIP             bool
 	DisableIPv6           bool
 }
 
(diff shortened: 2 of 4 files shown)


apricote:
Hey @Blitss, thanks for the PR! I am currently on vacation, and will take a look at this in January.

PReimers:
Any news here?

Blitss:
@jooola do you want me to resolve that changes for you or will you check it out?

jooola:
@Blitss Sorry for the delay, we are currently low resource to work on this Pull Request. But we don't want this PR to vanish without giving you a proper review/answer, I will therefor pin the PR for now.

No, don't bother about the conflict for now.

hcloud/instances_util.go
```diff
@@ -91,6 +92,7 @@ func getRobotServerByID(c robot.Client, id int, node *corev1.Node) (*hrobotmodel
 
 	// check whether name matches - otherwise this server does not belong to the respective node anymore
 	if server.Name != node.Name {
+		klog.Warningf("%s: server %d has name %q, but node %q has name %q. if you want node to be matched with node in Hetzner Robot you should rename it.", op, id, server.Name, node.Name, node.Name)
```
jooola, line 95: This seems unrelated to this PR, but might be a good addition.

internal/config/config.go
```diff
@@ -31,6 +31,7 @@ const (
 	hcloudLoadBalancersNetworkZone           = "HCLOUD_LOAD_BALANCERS_NETWORK_ZONE"
 	hcloudLoadBalancersDisablePrivateIngress = "HCLOUD_LOAD_BALANCERS_DISABLE_PRIVATE_INGRESS"
 	hcloudLoadBalancersUsePrivateIP          = "HCLOUD_LOAD_BALANCERS_USE_PRIVATE_IP"
+	hcloudLoadBalancersUseNodeIP             = "HCLOUD_LOAD_BALANCERS_USE_NODE_IP"
```
jooola, line 34: I am not sure about the naming or the option, this seems to target robot server only right?

jooola (commented):
I did a very quick review, but I'll not be able to give the final word on this PR.

hcloud/instances_util.go
Blitss: Do you want this to be a separate PR or just a separate commit?

paprickar:
Thanks @blitss for the work!
I need it now - and took your work https://github.com/paprickar/hcloud-cloud-controller-manager/tree/pr-590 until its merged.

FYI:
I most likely I will add a docker build job to keep track of upstream changes and run the modified version.

lukasmetzner:
Hey,

in our new version v1.24.0 we have made it possible to configure an additional InternalIP for Robot nodes via the kubelet flag `--node-ip`. These IPs can also be used in the context of Load Balancers when using the annotation `load-balancer.hetzner.cloud/use-private-ip: true`.

Thank you for your contribution ^^

I will go ahead and close this PR now.

Best Regards
Lukas

PR #597, opened 2024-01-08
feat: drop tests for kubernetes v1.25

apricote:
Kubernetes v1.25 is End of Life since 2023-10-28. In accordance with our Versioning policy.


apricote:
Follow up PR will add tests for 1.29.

jooola (approved):

PR #598, opened 2024-01-08
feat: emit Kubernetes events for error conditions

apricote:
`k/cloud-provider` emits Events for most errors we return. In some conditions we previously hid the errors from `k/cloud-provider` for various reasons and only logged a warning. This is hard to discover for users. With this change, we now also emit Kubernetes Events for these cases.

### DIFF
--- a/internal/hcops/load_balancer.go
+++ b/internal/hcops/load_balancer.go
@@ -643,15 +643,23 @@ func (l *LoadBalancerOps) ReconcileHCLBTargets(
 				continue
 			}
 
-			klog.InfoS("remove target", "op", op, "service", svc.ObjectMeta.Name, "targetName", k8sNodes[id].Name)
+			// k8sNodes[id] can be nil if the node is currently being deleted.
+			var nodeName string
+			if node := k8sNodes[id]; node != nil {
+				nodeName = node.Name
+			} else {
+				nodeName = fmt.Sprintf("%d", id)
+			}
+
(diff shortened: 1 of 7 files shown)


internal/hcops/load_balancer.go
```diff
@@ -699,27 +717,29 @@ func (l *LoadBalancerOps) ReconcileHCLBTargets(
...
 			UsePrivateIP: &usePrivateIP,
 		}
 		a, _, err := l.LBClient.AddServerTarget(ctx, lb, opts)
 		if err != nil {
 			if hcloud.IsError(err, hcloud.ErrorCodeResourceLimitExceeded) {
-				klog.InfoS("resource limit exceeded", "err", err.Error(), "op", op, "service", svc.ObjectMeta.Name, "targetName", k8sNodeNames[id])
-				return false, nil
+				klog.InfoS("resource limit exceeded", "err", err.Error(), "op", op, "service", svc.ObjectMeta.Name, "targetName", node)
+				// Continue loop so that error is emitted for each node
+				continue
```
apricote, line 737: Please change this if block to call `emitMaxTargetsReachedError()` instead, same as line 778.

pauhull: Done in a8c5f99

apricote (commented):
LGTM from my side

jooola (approved):

PR #631, opened 2024-04-13
Fix managed certificate load balancer service

anjmao:
Looks like this is the reason why managed certificate is never attached to lb https service.


apricote:
Hey @anjmao,

the branch that you modified is used for uploaded certificates. The branch below is used for managed certificates (Lines 1044-1062 right now).

Can you open an issue to explain the problem you are having?

jooola:
Closing for inactivity

PR #652, opened 2024-05-29
feat: read HCLOUD_TOKEN from file

### LINKED ISSUE #595: Support secret injection with HashiCorp Vault
I would like to see support for HashiCorp Vault.
More specifically, the Vault Agent Injector, which can inject secrets from Vault into pods.

To do this, HCCM must either read the HCLOUD_TOKEN from a file or enable the sourcing of a file before startup (so that injecting environment variables via vault is possible, see https://developer.hashicorp.com/vault/docs/platform/k8s/injector/examples#environment-variable-example).

Another solution would be to use the [Vault Secrets Operator](https://github.com/hashicorp/vault-secrets-operator/), which synchronises Vault Secrets with Kubernetes Secrets. However, I think loading on pod startup is better (because it can use the pod service account. This can also be done with the VSO but this requires 2 CRDs for each secret that should be loaded (`VaultAuth` and `VaultStaticSecret`))

What do the developers and the community think about this change?

simonostendorf:
This allows the `HCLOUD_TOKEN` (and `ROBOT_USER` and `ROBOT_PASSWORD`) to be read from a file. This can be useful if the token is injected using secret injection (e.g. with the vault agent injector).

I tested the changes on my dev cluster. I cannot test the robot account changes, because I only use hcloud. But because they use the same function inside the code I think its working too :). If someone is interested in using this with the vault agent injector, I used the following helm values:

```yaml
image:
  repository: <custom-image-because-changes-are-not-released>
  tag: <custom-image-because-changes-are-not-released>
podAnnotations:
  vault.hashicorp.com/agent-inject: "true"
  vault.hashicorp.com/log-format: json
  vault.hashicorp.com/role: <your-vault-role-name>
  vault.hashicorp.com/secret-volume-path-token: /vault/secrets
  vault.hashicorp.com/agent-inject-file-token: token
  vault.hashicorp.com/agent-inject-secret-token: <your-vault-mount>/data/<your-vault-path>
  vault.hashicorp.com/agent-inject-template-token: |
    {{ with secret "<your-vault-mount>/data/<your-vault-path>" -}}
     {{ .Data.data.token }}
    {{- end }}
env:
  HCLOUD_TOKEN_FILE:
    value: "/vault/secrets/token"
  HCLOUD_TOKEN: null # must be set because helm results in using value and valueFrom and that results in an error
```

This change is inspired from [external-dns cloudflare provider](https://github.com/kubernetes-sigs/external-dns/blob/master/provider/cloudflare/cloudflare.go#L171). I requested the same change for the [csi-driver](https://github.com/hetznercloud/csi-driver/pull/617) to keep consistency in reading HCLOUD_TOKEN from file.

Closes #595

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -31,11 +31,13 @@ env:
 
   # You can also use a file to provide secrets to the hcloud-cloud-controller-manager.
   # This is currently possible for HCLOUD_TOKEN, ROBOT_USER, and ROBOT_PASSWORD.
-  # The value of the env var (e.g. HCLOUD_TOKEN) must start with "file:" to indicate that the value is a file path.
+  # Use the env var appended with _FILE (e.g. HCLOUD_TOKEN_FILE) and set the value to the file path that should be read
   # The file must be provided externally (e.g. via secret injection).
   # Example:
-  # HCLOUD_TOKEN:
-  #   value: "file:/etc/hetzner/token"
+  # HCLOUD_TOKEN_FILE:
+  #   value: "/etc/hetzner/token"
+  # to disable reading the token from the secret you have to disable the original env var:
+  # HCLOUD_TOKEN: null
 
   HCLOUD_TOKEN:
     valueFrom:
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -29,6 +29,14 @@ env:
   # HCLOUD_NETWORK - see networking.enabled
   # ROBOT_ENABLED - see robot.enabled
 
+  # You can also use a file to provide secrets to the hcloud-cloud-controller-manager.
+  # This is currently possible for HCLOUD_TOKEN, ROBOT_USER, and ROBOT_PASSWORD.
+  # The value of the env var (e.g. HCLOUD_TOKEN) must start with "file:" to indicate that the value is a file path.
+  # The file must be provided externally (e.g. via secret injection).
+  # Example:
+  # HCLOUD_TOKEN:
+  #   value: "file:/etc/hetzner/token"
+
   HCLOUD_TOKEN:
     valueFrom:
       secretKeyRef:
@@ -103,3 +111,7 @@ nodeSelector: {}
 robot:
   # Set to true to enable support for Robot (Dedicated) servers.
   enabled: false
+
+podLabels: {}
+
+podAnnotations: {}
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -102,7 +102,7 @@ func readFromEnvOrFile(envVar string) (string, error) {
 	if strings.HasPrefix(value, "file:") {
 		valueBytes, err := os.ReadFile(strings.TrimPrefix(value, "file:"))
 		if err != nil {
-			return "", errors.New("failed to read " + envVar + " from file: " + err.Error())
+			return "", fmt.Errorf("failed to read %s from file: %w", envVar, err) 
 		}
 		return strings.TrimSpace(string(valueBytes)), nil
 	}
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -97,16 +97,30 @@ type HCCMConfiguration struct {
 	Route        RouteConfiguration
 }
 
+// read values from environment variables or from file set via _FILE env var
+// values set directly via env var take precedence over values set via file
 func readFromEnvOrFile(envVar string) (string, error) {
-	value := os.Getenv(envVar)
-	if strings.HasPrefix(value, "file:") {
-		valueBytes, err := os.ReadFile(strings.TrimPrefix(value, "file:"))
-		if err != nil {
-			return "", fmt.Errorf("failed to read %s from file: %w", envVar, err) 
-		}
-		return strings.TrimSpace(string(valueBytes)), nil
+	// check if the value is set directly via env (e.g. HCLOUD_TOKEN)
+	value, ok := os.LookupEnv(envVar)
+	if ok {
+		return value, nil
 	}
-	return value, nil
+
+	// check if the value is set via a file (e.g. HCLOUD_TOKEN_FILE)
+	value, ok = os.LookupEnv(envVar + "_FILE")
+	if !ok {
+		// return no error here, the values could be optional
+		// and the function "Validate()" below checks that all required variables are set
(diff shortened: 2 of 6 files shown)


internal/config/config.go
```diff
@@ -96,6 +97,18 @@ type HCCMConfiguration struct {
 	Route        RouteConfiguration
 }
 
+func readFromEnvOrFile(envVar string) (string, error) {
+	value := os.Getenv(envVar)
+	if strings.HasPrefix(value, "file:") {
+		valueBytes, err := os.ReadFile(strings.TrimPrefix(value, "file:"))
+		if err != nil {
+			return "", errors.New("failed to read " + envVar + " from file: " + err.Error())
```
jooola, line 105:
```suggestion
			return "", fmt.Errorf("failed to read %s from file: %w", envVar, err) 
```

I'd recommend wrapping the underlying error, and leverage `fmt.Errorf`.

internal/config/config_test.go
```diff
@@ -48,6 +48,25 @@ func TestRead(t *testing.T) {
...
+				Robot:        RobotConfiguration{User: "", Password: "", CacheTimeout: 0},
+				Metrics:      MetricsConfiguration{Enabled: false},
+				Instance:     InstanceConfiguration{},
+				LoadBalancer: LoadBalancerConfiguration{Enabled: false},
+				Route:        RouteConfiguration{Enabled: false},
+			},
+			wantErr: errors.New(`failed to read HCLOUD_TOKEN from file: open /etc/hetzner/token: no such file or directory
+failed to read ROBOT_USER from file: open /etc/hetzner/user: no such file or directory
+failed to read ROBOT_PASSWORD from file: open /etc/hetzner/password: no such file or directory`),
+		},
```
jooola, line 69: Could we test the happy path as well?

jooola (commented):
Overall, the feature looks great, I only have a few nits.

We will have to wait for @apricote for the final merge decision.

chart/values.yaml
```diff
@@ -29,6 +29,14 @@ env:
   # HCLOUD_NETWORK - see networking.enabled
   # ROBOT_ENABLED - see robot.enabled
 
+  # You can also use a file to provide secrets to the hcloud-cloud-controller-manager.
+  # This is currently possible for HCLOUD_TOKEN, ROBOT_USER, and ROBOT_PASSWORD.
+  # The value of the env var (e.g. HCLOUD_TOKEN) must start with "file:" to indicate that the value is a file path.
+  # The file must be provided externally (e.g. via secret injection).
+  # Example:
+  # HCLOUD_TOKEN:
+  #   value: "file:/etc/hetzner/token"
```
jooola, line 38: Have you considered using a new env var name ? For example, using `HCLOUD_TOKEN_FILE` instead of using a `file:` prefix in the exiting variable? What are the pros and cons?

internal/config/config_test.go
simonostendorf: I have not yet covered the positive case in the test cases, as I do not know how you deal with file creation in tests. To run the test for the positive case, a file (e.g. `/tmp/hcloud/token`) would have to be created first.

chart/values.yaml
simonostendorf: Yes, I have thought about whether a new variable makes sense. I decided to use HCLOUD_TOKEN because it has been used everywhere and is therefore familiar to users. For the same reason, I can understand why you would want to use HCLOUD_TOKEN instead of HCLOUD_TOKEN_FILE. However, in my opinion, HCLOUD_TOKEN should be preferred, as it is currently the standard at Hetzner. 

You could also handle the whole thing like AWS, OpenStack etc, which have a standard location for credential files, but then this would still need to be defined, documented and built into as many Hetzner products as possible (e.g. /etc/hetzner/credentials, which could then contain the HCLOUD_TOKEN, ROBOT_USER, ROBOT_PASSWORD and perhaps other credentials as a JSON file), but I think this is not in the scope of this PR.

internal/config/config_test.go
simonostendorf: Maybe `internal/testsupport` could be appended to allow tmp file creation.

chart/values.yaml
apricote: Having a default discovery of the token+context that is used by all tools would be my dream, but we do not have that right now.

---

I prefer a separate variable `HCLOUD_TOKEN_FILE`. `HCLOUD_TOKEN` is already standardized across many tools as being the (plain text) token. Adding another format that is only supported in a single (or 2) tool adds to the confusion.

---

I think the prefix should be `file:/` to match URIs that already support file paths like `file://home/julian/` (ie. as implemented in browsers). Though that is not relevant if specified in a `_FILE` env variable.

simonostendorf: Thank you for your input, i will add a `HCLOUD_TOKEN_FILE` variable that contains the file path with the prefix `file:/`.

So an example value would be `HETZNER_TOKEN_FILE=file://etc/hetzner/token`

apricote: I think in that case we can drop the `file:/` prefix

internal/testsupport/files.go
```diff
@@ -0,0 +1,40 @@
+package testsupport
+
+import (
+	"os"
+	"testing"
+)
+
+func Setfiles(t *testing.T, files map[string]string) func() {
```
apricote, line 8:
```suggestion
// SetFiles can be used to temporarily create files on the local file system.
// It returns a function that will clean up all files it created.
func SetFiles(t *testing.T, files map[string]string) func() {
```

What do you think about creating the files in `/tmp` (not sure about Windows :thinking:), to make sure that they get cleaned up one way or another?

apricote:
Thanks for the PR @simonostendorf!

Dont worry about the `e2e tests` failing, they dont work for people outside of the `hetznercloud` organization on GitHub.

chart/values.yaml
simonostendorf: Switched to `_FILE` env vars in newest commit.

apricote (approved):
Thank you very much @simonostendorf! I pushed a small commit with two style fixes. The e2e tests also were successful locally. Going to merge this now :)

PR #667, opened 2024-06-19
docs: Always use latest flannel release

madalinignisca:
This will always use the latest flannel release, not the main branch. Pointing to an exact version would keep the documentation to a version that might not work for users in the future.

The snippet is from Flannel's README.


apricote (approved):
Thanks :rocket:

PR #670, opened 2024-06-27
chore: reuse robot development server using overlayfs

jooola:
- Install the robot server once a week. Use `overlayfs` to allow resetting the whole system with a reboot.
- Reboot (and reset) the robot server before starting the e2e tests.


hack/robot-e2e/install.yml
```diff
@@ -0,0 +1,123 @@
...
+      ansible.builtin.apt:
+        autoremove: true
+        purge: true
+        state: absent
+
+    - name: Deploy overlayroot config
+      ansible.builtin.copy:
+        content: |
+          overlayroot=tmpfs
+        dest: /etc/overlayroot.conf
```
apricote, line 116: Lovely if the solution to ones problem turns out to be a single config file :D

hack/robot-e2e/reset.yml
```diff
@@ -0,0 +1,35 @@
...
+      delegate_to: localhost
+      ansible.builtin.command: >-
+        k3sup join
+          --server-ip='{{ lookup('ansible.builtin.env', 'CONTROL_IP') }}'
+          --ip='{{ ansible_host }}'
+          --k3s-channel='stable'
+          --k3s-extra-args="
+            --kubelet-arg cloud-provider=external
+            --node-label instance.hetzner.cloud/is-root-server=true"
+          --ssh-key ./ssh_key
```
apricote, line 35: This ssh key is used to connect to the `--server-ip` (to get the token) and to the `--ip` (to bootstrap the node).

Maybe you can add the key from `hack/.ssh-$SCOPE` to the server in a step before and use that instead?

apricote (commented):
Looking very good already :)

apricote:
Now included in #671.

PR #671, opened 2024-06-28
refactor: new dev/testing environment

apricote:


.github/workflows/test_e2e.yml
```diff
@@ -33,105 +41,100 @@ jobs:
...
             --output yaml \
-          > debug-logs/events.yaml
-
-      - name: Show HCCM Logs on Failure
-        if: failure()
-        continue-on-error: true
-        run: |
-          echo "::group::hccm.log"
-          cat debug-logs/hccm.log
+          | tee debug-logs/events.yaml
```
jooola, line 106:
```suggestion
            | tee debug-logs/events.yaml
```

.github/workflows/test_e2e.yml
```diff
@@ -33,105 +41,100 @@ jobs:
...
         continue-on-error: true
         uses: actions/upload-artifact@v4
         with:
-          name: debug-logs-${{ env.SCOPE }}
+          name: debug-logs-${{ env.ENV }}
           path: debug-logs/
 
   robot:
     name: Robot
+
```
jooola, line 124:
```suggestion
```

.github/workflows/test_e2e.yml
```diff
@@ -33,105 +41,100 @@ jobs:
...
-    # Make sure that only one Job is using the server at a time
-    concurrency: robot-test-server
+    # Make sure that only one job is using the server at a time
+    concurrency: robot
     environment: e2e-robot
 
     env:
-      K3S_VERSION: v1.29.0+k3s1
-      SCOPE: gha-${{ github.run_id }}-${{ github.run_attempt }}-robot
+      K3S_CHANNEL: v1.29
```
jooola, line 133: Should we add the renovate regex manager?

.github/workflows/test_e2e.yml
```diff
@@ -143,102 +146,89 @@ jobs:
...
-          echo "CONTROL_IP=$CONTROL_IP" >> "$GITHUB_ENV"
-
-      - name: Build and Deploy HCCM
-        run: |
-          skaffold build --tag="e2e-${GITHUB_RUN_ID}-${GITHUB_RUN_NUMBER}"
-          tag=$(skaffold build --tag="e2e-${GITHUB_RUN_ID}-${GITHUB_RUN_NUMBER}" --quiet --output="{{ (index .Builds 0).Tag }}")
-          skaffold deploy \
-            --profile=robot \
-            --images=hetznercloud/hcloud-cloud-controller-manager=$tag
+          mkdir ~/.ssh # required for ssh-copy-id
```
jooola, line 178:
```suggestion
          mkdir -p ~/.ssh # required for ssh-copy-id
```

.github/workflows/test_e2e.yml
```diff
@@ -143,102 +146,89 @@ jobs:
...
+          install --mode=600 <(echo -n "$ROBOT_SSH_KEY") ./ssh_key
+          eval "$(ssh-agent -s)"
+          ssh-add ./ssh_key
+          trap "rm -f ./ssh_key; kill $SSH_AGENT_PID" EXIT
```
jooola, line 182: Might be worth having a small wrapper since we use this in multiple places, e.g. `with-ssh-agent.sh`, that wraps the setup and teardown, and any command can be passed as argument.
 
That we could call like this: 
```
with-ssh-agent.sh make -C dev up
```

.github/workflows/test_e2e.yml
```diff
@@ -143,102 +146,89 @@ jobs:
...
             --output yaml \
-          > debug-logs/events.yaml
-
-      - name: Show HCCM Logs on Failure
-        if: failure()
-        continue-on-error: true
-        run: |
-          echo "::group::hccm.log"
-          cat debug-logs/hccm.log
+          | tee debug-logs/events.yaml
```
jooola, line 220:
```suggestion
            | tee debug-logs/events.yaml
```

README.md
```diff
@@ -200,132 +200,100 @@ Current Kubernetes Releases: https://kubernetes.io/releases/
...
-SKAFFOLD_DEFAULT_REPO=your_docker_hub_username skaffold dev
+If you want to work on the Robot support, you need to make some changes to the above setup.
+
+This requires that you have a Robot Server in the same account you use for the development. The server needs to be setup with the Ansible Playbook `dev/robot/install.yml` and configured in `dev/robot/install.yml`.
+
+1. Set these environment variables:
+
+```shell
+export TF_VAR_setup_robot=true
+export ROBOT_ENABLED=true
```
jooola, line 284: Could we deduplicate this ? We should move the ROBOT_ENABLED var in the makefile, and write the env.autovars file during the setup.

dev/Makefile
```diff
@@ -0,0 +1,33 @@
+SHELL = bash
+.SHELLFLAGS = -e -c
+.ONESHELL:
+
+ENV ?= dev
+K3S_CHANNEL ?= stable
+
```
jooola, line 7:
```suggestion
ENV ?= dev
K3S_CHANNEL ?= stable
ROBOT_ENABLED ?= false
```

dev/Makefile
```diff
@@ -0,0 +1,33 @@
...
+env.auto.tfvars:
+	@echo 'name           = "$(ENV)"' > "$@"
+	@echo 'hcloud_token   = "$(HCLOUD_TOKEN)"' >> "$@"
+	@echo 'k3s_channel    = "$(K3S_CHANNEL)"' >> "$@"
+	@echo 'robot_user     = "$(ROBOT_USER)"' >> "$@"
+	@echo 'robot_password = "$(ROBOT_PASSWORD)"' >> "$@"
+
```
jooola, line 14:
```suggestion
	@echo 'robot_enabled = "$(ROBOT_ENABLED >> "$@"
	@echo 'robot_user     = "$(ROBOT_USER)"' >> "$@"
	@echo 'robot_password = "$(ROBOT_PASSWORD)"' >> "$@"

```

Formatting is broken in the suggestion, please adjust accordingly :3

dev/main.tf
```diff
@@ -0,0 +1,13 @@
+module "dev" {
+  source = "github.com/hetznercloud/terraform-k8s-dev?ref=v0.3.0"
```
jooola, line 2:
```suggestion
  source = "github.com/hetznercloud/kubernetes-dev-env?ref=v0.3.0"
```

dev/main.tf
```diff
@@ -0,0 +1,13 @@
...
+
+  name         = "hccm-${replace(var.name, "/[^a-zA-Z0-9-_]/", "-")}"
+  worker_count = 1
+  # We deploy hccm through skaffold, its the application under development/test.
+  deploy_hccm      = false
+  use_cloud_routes = !var.setup_robot
```
jooola, line 8: Should we use a consistent naming? 

- `deploy_hccm` => `hccm_enabled`
- `setup_robot` => `robot_enabled`

dev/robot.tf
```diff
@@ -0,0 +1,105 @@
...
+        root@${local.robot_ipv4}
+    EOT
+  }
+}
+
+module "registry_robot" {
+  count      = var.setup_robot ? 1 : 0
+  depends_on = [null_resource.reset_robot]
+
+  source = "github.com/hetznercloud/terraform-k8s-dev//k3s_registry?ref=v0.3.0"
```
jooola, line 49:
```suggestion
  source = "github.com/hetznercloud/kubernetes-dev-env//k3s_registry?ref=v0.3.0"
```

Are the double `//` expected?

dev/robot.tf
```diff
@@ -0,0 +1,105 @@
...
+      k3sup join \
+        --ssh-key='${module.dev.ssh_private_key_filename}' \
+        --ip='${local.robot_ipv4}' \
+        --server-ip='${module.dev.control_server_ipv4}' \
+        --k3s-channel='${var.k3s_channel}' \
+        --k3s-extra-args="\
+          --kubelet-arg='cloud-provider=external' \
+          --node-ip='${local.robot_ipv4}' \
+          --node-label instance.hetzner.cloud/is-root-server=true \
+          --snapshotter=native" \
```
jooola, line 79: Could we add a comment about why we use this snapshotter please?

dev/robot.tf
```diff
@@ -0,0 +1,105 @@
...
+      EOT
+  }
+}
+
+data "local_sensitive_file" "kubeconfig" {
+  depends_on = [module.dev]
+  filename   = "${path.root}/files/kubeconfig.yaml"
```
jooola, line 86: Should this also be an output in the terraform module?

dev/variables.tf
```diff
@@ -0,0 +1,23 @@
...
+  sensitive = true
+}
+variable "k3s_channel" {
+  type = string
+}
+variable "setup_robot" {
+  type    = bool
+  default = false
+}
+# Hetzner Robot
```
jooola, line 15:
```suggestion
# Hetzner Robot
variable "robot_enabled" {
  type    = bool
  default = false
}
```

dev/robot/install.yml
```diff
@@ -0,0 +1,123 @@
...
+
+  tasks:
+    - name: Upload CI SSH Key # noqa args[module]
+      delegate_to: localhost
+      community.hrobot.ssh_key:
+        name: hcloud-cloud-controller-manager
+        public_key: ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIA/+QBJNjrtqJjdsOqOwfELUS+3M3g0QRbn7pQaXfw7f ci@2024-06-27
+        state: present
+      register: ci_ssh_key
+
```
jooola, line 25:
```suggestion

    - name: Set server name # noqa args[module]
      community.hrobot.server:
        server_number: "{{ server_number }}"
        server_name: "{{ inventory_hostname }}"

```
We fixed it manually, but I think we should have a task to rename the robot server to the `inventory_hostname`:

jooola (changes requested):
Looks great, just a few changes left.

.github/workflows/test_e2e.yml
jooola: I'll push a script.

jooola: The new script was pushed, please update your local branch ;)

jooola:
We are also missing a reset triggered at the end of the pipeline. Happy to only implement it in a later PR.

.github/workflows/test_e2e.yml
apricote: To pin a specific version? We need to bump the channels (in non-robot matrix) & docs anyway when we upgrade, so I dont see the benefit here.

.github/workflows/test_e2e.yml
apricote: Thanks!

dev/robot.tf
apricote: Yup, the double `//` is for "directory inside of the git repo": https://developer.hashicorp.com/terraform/language/modules/sources#modules-in-package-sub-directories

dev/robot.tf
apricote: https://github.com/hetznercloud/kubernetes-dev-env/pull/22

dev/robot/install.yml
apricote: I would prefer to avoid meddling with the server name in Robot. Feels dirty. What do you think about inverting this and using the actual server name as the hostname instead?

```patch
diff --git a/dev/robot/install.yml b/dev/robot/install.yml
index ec191c44..598dcbd7 100644
--- a/dev/robot/install.yml
+++ b/dev/robot/install.yml
@@ -47,10 +47,15 @@
   hosts: all
   gather_facts: true
   tasks:
+    - name: Get hostname
+      community.hrobot.server_info:
+        server_number: "{{ server_number }}"
+      register: server_result
+
     - name: Deploy installimage autosetup
       ansible.builtin.copy:
         content: |
-          HOSTNAME {{ inventory_hostname }}
+          HOSTNAME {{ server_result.servers[0].server_name }}
 
           DRIVE1 /dev/sda
           DRIVE2 /dev/sdb
```

jooola: This would more be the role of an inventory plugin. Let's keep it as it is, and see in the future if we need it.

jooola (approved):

PR #674, opened 2024-07-02
refactor(hcops): cleanup AllServersCache

apricote:
This separates accessing the cache from loading the data and renames some identifiers to better represent their purpose.

### DIFF
--- a/internal/hcops/server.go
+++ b/internal/hcops/server.go
@@ -16,8 +16,8 @@ import (
 // AllServersCache caches the result of the LoadFunc and provides random access
 // to servers using select hcloud.Server attributes.
 //
-// To simplify things the allServersCache reloads all servers on every cache
-// miss, or whenever a timeout expired.
+// To simplify things the AllServersCache reloads all servers on every cache
+// miss, or whenever the data is older than MaxAge.
 type AllServersCache struct {
 	LoadFunc    func(context.Context) ([]*hcloud.Server, error)
 	LoadTimeout time.Duration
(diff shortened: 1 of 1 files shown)


internal/hcops/server.go
```diff
@@ -74,39 +74,69 @@ func (c *AllServersCache) ByName(name string) (*hcloud.Server, error) {
...
+		if err := c.refreshCache(); err != nil {
+			return nil, fmt.Errorf("%s: %w", op, err)
+		}
+
+		// Check once again if the server was found
+		server, ok = retrieveFromCacheMaps()
+		if ok {
+			return server, nil
+		}
 	}
```
apricote, line 111: This is missing coverage right now, will be added in a follow up MR.

internal/hcops/server.go
```diff
@@ -117,22 +147,18 @@ func (c *AllServersCache) getCache(getSrv func() (*hcloud.Server, bool)) (*hclou
...
-				klog.Warningf("Overriding AllServersCache entry for private ip %s with server %s\n", n.IP.String(), srv.Name)
+				klog.Warningf("Overriding AllServersCache entry for private ip %s with server %s\n", n.IP.String(), server.Name)
 			}
-			c.byPrivIP[n.IP.String()] = srv
+			c.byPrivIP[n.IP.String()] = server
 		}
 
 		// Index servers by their names.
-		c.byName[srv.Name] = srv
+		c.byName[server.Name] = server
```
jooola, line 156:
```suggestion
		c.byName[server.Name] = servers[i]
```

Is there a chance that are storing the loop reference every time?  Shouldn't we prefer the indexed loop to make it explicit?

apricote: I do not think we were affected by this, otherwise the whole thing would not work at all. But we do not have to care about this anymore, the language was changed in Go 1.22 to avoid this issue altogether: https://go.dev/doc/go1.22#language

jooola (approved):

Issue #395, opened 2023-03-15
fix: many API requests from Routes Controller


apricote:
We are sending a request every ten seconds to Hetzner Cloud API if the routing functionality is active (default when Networking is activated). This equals 1/10th of the default API rate limit on new accounts.

![Example metrics that demonstrate the issue described here](https://user-images.githubusercontent.com/3393079/225241350-264ab8a7-b95b-4134-86d4-9ce548aab303.png)

This happens even though nothing has changed on the Kubernetes side, and there is no real need to reconcile this often. It should be enough to reconcile this on Node IP changes and when Nodes are added or removed.

This behaviour is coming from [`k/cloud-provider`](https://github.com/kubernetes/cloud-provider/blob/0dac8cc4d471bfbfdf48e7d6a9c039efb8d21d88/controllers/route/route_controller.go#L113-L122) and there exists an upstream issue for it: kubernetes/kubernetes#60646

samcday:
The main issue here is the fact that [route reconciliation isn't triggered properly](https://github.com/kubernetes/cloud-provider/blob/0dac8cc4d471bfbfdf48e7d6a9c039efb8d21d88/controllers/route/route_controller.go#L113-L117).

If the reconciliation logic was more internally coherent, we could safely increase the default 10s re-reconciliation interval to something like 1m. This would mean that new servers should get routes quickly, but routes that were modified out of band might take a minute or two to self-heal.

Longer term we've been discussing the possibility/merits of implementing hccm's logic in a standard kubebuilder project. In that case we would have full control over things like this and be able to much more easily tune this without resorting to forking the upstream or pinging a 5 year old ticket.

In the interim, I propose configuring the [reconciliation period](https://github.com/kubernetes/cloud-provider/blob/0dac8cc4d471bfbfdf48e7d6a9c039efb8d21d88/config/v1alpha1/defaults.go#L69) from 10s -> 30s. This means that new servers take a little longer to get set up, but the normal case is 3x less API traffic.

apricote:
> In the interim, I propose configuring the [reconciliation period](https://github.com/kubernetes/cloud-provider/blob/0dac8cc4d471bfbfdf48e7d6a9c039efb8d21d88/config/v1alpha1/defaults.go#L69) from 10s -> 30s. This means that new servers take a little longer to get set up, but the normal case is 3x less API traffic.

Implemented in #403.

jooola:
[...]

PR #675, opened 2024-07-02
fix(routes): many requests for outdated routes by rate limiting

### LINKED ISSUE #673: Too many API requests for outdated routes in the network
### TL;DR

A  lot of calls to `client.Server.All()` are being made when the network has a lot of "outdated" routes for IPs that no longer belong to an active server.

### Expected behavior

HCCM should not spam the API in reasonable situation.

### Observed behavior

In `RouteController.ListRoutes()` we try to match every route in the network to the correct Kubernetes node.

For this, we check if the Gateway IP of the route matches the known internal IP of the any server in the network.

To make this more efficient, there was a cache introduced that provides access to the server list. This works great when the server is actually in the cache.

If no server can be found in the cache, we assume that our cache is outdated and refresh the cache with the API.

This refresh is repeated once for every route that has no matching server.

### Minimal working example

- Add x bogus routes in the Cluster CIDR to the Network before starting HCCM.
- Observe x calls to `GET /v1/servers`

### Log output

_No response_

### Additional information

_No response_

apricote:
Fixes #673

In `routes.ListRoutes()` we have to find the matching server/node for every route in the network. We find the server by utilizing a cache that maps every private IP to the corresponding server.

This cache has a feature that refreshes the list of servers if an entry can not be found. This is sensible, as the server might have been just created. This is also fatal, as this refresh happens for every single cache access. If there are a lot of routes in the network that do not belong to any server, we refresh the cache many times for each `ListRoutes()`. This is even more serious, as `ListRoutes()` is being called every 10-30 seconds (see #395).

This commit introduces a `rate.Limiter` in the `AllServersCache` which only allows the refresh to happen every 30 seconds.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.9.0
 	github.com/syself/hrobot-go v0.2.5
+	golang.org/x/time v0.3.0
 	k8s.io/api v0.30.2
 	k8s.io/apimachinery v0.30.2
 	k8s.io/client-go v0.30.2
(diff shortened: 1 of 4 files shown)


jooola (approved):
I am a bit afraid of the unknown side effects of this workaround. I think we can merge this, but we might need to come up with a better fix in the future.

jooola:
I rebased onto main to resolve the go.mod conflict.

apricote:
The primary side effect here is that new nodes might need additional 30s until routes are setup, if the project already had wonky routes.

PR #685, opened 2024-07-05
chore: skaffold was not using the built image

apricote:
We changed the image reference in a recent PR. With this, skaffold was unable to match it against the image referenced in the Helm chart, so it always deployed the tagged version of the image.

This caused all e2e tests to not use the actual changes made in those PRs.

Noticed this because the release-please PR was trying to pull a `v1.20.0` image and failed.

### DIFF
--- a/skaffold.yaml
+++ b/skaffold.yaml
@@ -4,7 +4,7 @@ metadata:
   name: cloud-controller-manager
 build:
   artifacts:
-    - image: docker.io/hetznercloud/hcloud-cloud-controller-manager
+    - image: hetznercloud/hcloud-cloud-controller-manager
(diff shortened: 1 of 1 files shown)


jooola (approved):

jooola:
Should we also update on the csi-driver side https://github.com/hetznercloud/csi-driver/blob/main/skaffold.yaml#L7 ?

apricote:
csi-driver should be fine, it references the images with that prefix in the Helm Chart: https://github.com/hetznercloud/csi-driver/blob/4c77198f9138dfc672cedea184551de2c20857b0/chart/values.yaml#L131

PR #686, opened 2024-07-08
feat(helm): allow setting affinity for deployment

apricote:
Support setting the Affinity for pods of the `Deployment`.


jooola (approved):
I assume you tested it? No need for a snapshot?

apricote:
Yup, tested it locally. Its the exact same code as for the `nodeSelector` right above it. Also tested by #687.

We do not have helm snapshots tests for this repo yet.

apricote:
Added an issue for the snapshots: https://github.com/hetznercloud/hcloud-cloud-controller-manager/issues/688

PR #692, opened 2024-07-08
chore(dev): speed up image build with ko

apricote:
Ko builds the binary locally and utilizes the local go build cache.

This makes the deployment with skaffold faster locally and in our e2e tests.

This makes the dev setup more different than the production setup (still using Dockerfile). But they were already a bit different anyway. In the mid-term I want to use `ko` for the production images too.


apricote:
In e2e tests this reduced the time of the `skaffold run` step from ~1m50s to ~1m30s (-18%).

lukasmetzner:
Let's close this one for now and tackle it in a v2. We can keep the branch around :)

Issue #118, opened 2020-11-13
Service of type LB gets private and public external-ip


plicon:
I've setup the cloud controller and configured it to automagically create a hetzner LB, so far so good. The load balancer is connected to the same LAN as the k8s worker nodes. 
I setup a service using the following definition, which nicely also creates the LB. 

```yaml
apiVersion: v1
kind: Service
metadata:
  name: nginx
  annotations:
    load-balancer.hetzner.cloud/location: hel1
    load-balancer.hetzner.cloud/use-private-ip: "true"
spec:
  selector:
    app: nginx
  type: LoadBalancer
  ports:
    - protocol: TCP
      port: 80
      targetPort: 80
```

When i describe the created the service it shows both the LB's private and PUBLIC ip-adress as shown below: 
```yaml
Name:                     nginx
Namespace:                traefik
Labels:                   <none>
Annotations:    
                          load-balancer.hetzner.cloud/location: hel1
                          load-balancer.hetzner.cloud/use-private-ip: true                         
                          field.cattle.io/publicEndpoints: [{"addresses":["10.0.10.4","95.217.175.215"],"port":80,"protocol":"TCP","serviceName":"traefik:nginx","allNodes":false}]
Selector:                 app=nginx
Type:                     LoadBalancer
IP:                       10.43.240.86
LoadBalancer Ingress:     95.217.175.215, 10.0.10.4
Port:                     <unset>  80/TCP
TargetPort:               80/TCP
NodePort:                 <unset>  30257/TCP
Endpoints:                10.244.2.23:80
Session Affinity:         None
External Traffic Policy:  Cluster
Events:                   <none>
```

The issue I then ran into is that using [external-dns](https://github.com/kubernetes-sigs/external-dns) with cloudflare for both ip-adresses a dns a-record is created. 

I would expect the service to only have the public ip-adress as loadbalacner-ingress.

0SkillAllLuck:
Hello @plicon, you can actually specify which ip-address is being used by external-dns: [How do I specify that I want the DNS record to point to either the Node's public or private IP when it has both?](https://github.com/kubernetes-sigs/external-dns/blob/master/docs/faq.md#how-do-i-specify-that-i-want-the-dns-record-to-point-to-either-the-nodes-public-or-private-ip-when-it-has-both). I hope this will help you.

plicon:
[...]

PR #49, opened 2020-06-23
Hetzner Cloud Load Balancer support

fhofherr:
Closes #48

PR #715, opened 2024-08-13
fix: populate ingress private ip when disable-private-ingress is false

### LINKED ISSUE #553: LoadBalancer does not populate IP field when `disable-public-network` is set.
I have created a service of type `LoadBalancer`, with the annotation `load-balancer.hetzner.cloud/disable-public-network: "true"`. The load-balancer has a private IP address assigned on my cluster's network (`172.29.0.3`), and responds correctly on this address. But, the status does not contain the LB's IP address:

```yaml
  status:
    loadBalancer: {}
```

When I create a service without the annotation, the status is correctly updated:

```yaml
  status:
    loadBalancer:
      ingress:
      - ip: 167.235.<snip>
      - ip: 2a01:4f8:<snip>
```

I think even if public access is disabled, the external IP should still be populated with the service's private network address, so that servers on my HCloud network (but outside the k8s cluster) can extract the address they can reach the service on. Also, without the load-balancer IP address, `external-dns` fails to add a record for this load-balancer.

jooola:
Closes #553

Refactor getting the ingresses for a load-balancer for both get load-balancer and ensure load-balancer, into using a shared function. This ensures we are consistent between the 2 methods.

Release notes:
The load balancer ingress IPs are now populated with the private IPs, unless the `load-balancer.hetzner.cloud/disable-private-ingress` annotation is set to `true`. Please make sure that you configured the annotation according to your needs, for example if you are using `external-dns`.

### DIFF
--- a/hcloud/load_balancers.go
+++ b/hcloud/load_balancers.go
@@ -86,23 +86,12 @@ func (l *loadBalancers) GetLoadBalancer(
 		}, true, nil
 	}
 
-	ingresses := []corev1.LoadBalancerIngress{
-		{
-			IP: lb.PublicNet.IPv4.IP.String(),
-		},
-	}
-
-	disableIPV6, err := l.getDisableIPv6(service)
+	ingress, err := l.buildLoadBalancerStatusIngress(lb, service)
 	if err != nil {
 		return nil, false, fmt.Errorf("%s: %v", op, err)
 	}
-	if !disableIPV6 {
-		ingresses = append(ingresses, corev1.LoadBalancerIngress{
-			IP: lb.PublicNet.IPv6.IP.String(),
-		})
-	}
 
-	return &corev1.LoadBalancerStatus{Ingress: ingresses}, true, nil
+	return &corev1.LoadBalancerStatus{Ingress: ingress}, true, nil
 }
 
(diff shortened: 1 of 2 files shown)


jooola:
image

apricote (approved):
Awesome!

apricote:
Ah, this now always adds the private ip to the addresses if set. IIRC we discussed that we want to add it only when no public IPs are available.

apricote (changes requested):
See above comment

jooola:
In that case, I am not sure to fully understand why it should be different between `EnsureLoadBalancer` and `GetLoadBalancer`. According to the cloud provider interface, I understand that both should not be really different.

If we need this difference to prevent a breaking change, we might not respect the `load-balancer.hetzner.cloud/disable-public-network` and  `load-balancer.hetzner.cloud/disable-private-ingress` annotations.

I think I understood that external dns might break because we are referencing a private IP that will never be reachable from the outside, but in that case people would have to set `disable-private-ingress`.

I might be missing something though :)

apricote:
We investigated the history of this today. Looking at #49 and #118 it was intended that we always add external+internal IPs the the Service Status. The annotation `load-balancer.hetzner.cloud/disable-private-ingress=true` was meant to help with setups where this causes issues.

So this PR fixes a bug and restores the intended behavior.

apricote (approved):

PR #729, opened 2024-08-22
fix: wrong version logged on startup

apricote:
The k/cloud-provider library is logging a version on startup. This version is expected to be set by Kubernetes build scripts, but we do not use this.

Right now it looks like this:

```
controllermanager.go:169] Version: v0.0.0-master+$Format:%H$
```

With the fix, this will report the same version as we use for the user-agent.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -23,6 +23,7 @@ builds:
       - -s
       - -w
       - -X github.com/hetznercloud/hcloud-cloud-controller-manager/hcloud.providerVersion={{ if not .IsSnapshot }}v{{ end }}{{ .Version }}
+      - -X k8s.io/component-base/version.gitVersion={{ if not .IsSnapshot }}v{{ end }}{{ .Version }}
 
 dockers:
   - build_flag_templates: [--platform=linux/amd64]


jooola:
Should we maybe add some more details ?

```
BUILD_DATE="$(date -u +'%Y-%m-%dT%H:%M:%SZ')"

-X k8s.io/client-go/pkg/version.gitVersion=${VERSION}
-X k8s.io/client-go/pkg/version.gitCommit=${COMMIT}
-X k8s.io/client-go/pkg/version.gitTreeState=${TREE_STATE}
-X k8s.io/client-go/pkg/version.buildDate=${BUILD_DATE}

-X k8s.io/component-base/version.gitVersion=${VERSION}
-X k8s.io/component-base/version.gitCommit=${COMMIT}
-X k8s.io/component-base/version.gitTreeState=${TREE_STATE}
-X k8s.io/component-base/version.buildDate=${BUILD_DATE}
```

apricote:
I am against that, build information is automatically added through https://pkg.go.dev/runtime/debug#ReadBuildInfo. 

I only patch this field because its being logged by a component outside of our control.

jooola (approved):

PR #732, opened 2024-08-26
Skip udp service ports as loadbalancers do not support UDP

lhp-nemlig:
If a loadbalancer svc includes both tcp and udp for the same port, for example 53/tcp and 53/udp for dns, the CCM will currently fail to create the loadbalancer due to duplicate ports, but the CCM should ignore the UDP port as loadbalancers do not support udp.

### DIFF
--- a/internal/hcops/load_balancer.go
+++ b/internal/hcops/load_balancer.go
@@ -842,6 +842,11 @@ func (l *LoadBalancerOps) ReconcileHCLBServices(
 		)
 
 		portNo := int(port.Port)
+		if port.Protocol == corev1.ProtocolUDP {
+			klog.InfoS("skipping port, load balancers do not support UDP", "op", op, "port", portNo, "loadBalancerID", lb.ID)
+			continue
+		}
+
 		portExists := hclbListenPorts[portNo]
 		delete(hclbListenPorts, portNo)
 


apricote:
Hey @lhp-nemlig, thanks for the pull request.

I disagree with this statement: "but the CCM should ignore the UDP port".

In my opinion, we should loudly fail the provide a Load Balancer for this service, so the user notices that the provided configuration does not work and provide an alternative.

We should validate that the users desired spec (the whole service) can be made available. If something of that desired state is not possible to implement, we should not hide this fact from the user. Instead the user should create a service that only has the TCP port if they want to use the Hetzner Cloud Load Balancer.

"Loudly failing" would mean:

- `Event` for the Service with the error
- `Condition` on the Service
- Log statement in `hcloud-cloud-controller-manager`

lhp-nemlig:
The alternative would work yes, but in my experience some helm charts do not have the possibility to disable their UDP svc port, but I see your point and I think you are right.
I guess i will solve my own issues with helm post-render.
Is it possible that you could push your above suggestions for the failure mode into this PR, my experience with this codebase is limited :)

lukasmetzner:
Opened new issue #799  for implementing the user feedback. Closing this PR.

PR #744, opened 2024-09-11
feat(helm): allow setting extra pod volumes via chart values 

### LINKED ISSUE #743: feat: allow extraVolumes and extraVolumeMounts in helm chart
### TL;DR

The user @simonostendorf  provided the great feature in #652 to read the `HCLOUD_TOKEN` from an file! It is great having the possibility to read the token from a file since leaking files is much less likely then envs.

Unfortunately I have a hard time using this feature when you don't use HashiCorps Secret injector. Normally you would mount the secret as a volume but you cant do this currently since the chart does not support any `extraVolumes` or `extraVolumeMounts`. But this is kind of needed when you use CSI Secret Driver or just want to mount the k8s secret as a file. 

### Expected behavior

Being able to set  `extraVolumes` or `extraVolumeMounts` so I can use my existing k8s secret and mount it as file via secret-key-ref oder to use a kubernetes secret csi driver definition as volume.

Is this something worth supporting? I would happily provide a PR if this is an accepted feature from Hetzner.

**Update:**
PR in #744

hegerdes:
## What changed
This adds the ability to set `extraVolumes` and `extraVolumeMounts` via helm to the pod. It can be used to mount extra files, to persist files inside the container or to allow usage of the CSI Secret Driver.   
It does not introduce user facing changes since the default for both is an empty array, so no volumes or volumeMounts are created when not explicitly set. 

Fixed #743 

##  Testing
The change can be previewed with the following command: `helm template chart -s templates/deployment.yaml --values chart/my-values.yaml --debug`

```yaml
# my-values.yaml 
# Mounts the specified volume to the hcloud-cloud-controller-manager container.
extraVolumeMounts:
  - name: token-volume
    readOnly: true
    mountPath: /var/run/secrets/hcloud

# Adds extra volumes to the pod.
extraVolumes:
  - name: token-volume
    secret:
      secretName: hcloud-token

# Set envs for token
env:
  HCLOUD_TOKEN_FILE:
    value: /var/run/secrets/hcloud/token
  HCLOUD_TOKEN: null
```

Testing can be done with the same file just with `helm install`. The `hcloud-token` must exist before deployment

### DIFF
--- a/chart/templates/daemonset.yaml
+++ b/chart/templates/daemonset.yaml
@@ -92,5 +92,13 @@ spec:
             {{- end }}
           resources:
             {{- toYaml $.Values.resources | nindent 12 }}
+          {{- with .Values.extraVolumeMounts }}
+          volumeMounts:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
       priorityClassName: system-cluster-critical
+      {{- with .Values.extraVolumes }}
+      volumes:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
 {{- end }}
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -100,5 +100,13 @@ spec:
             {{- end }}
           resources:
             {{- toYaml $.Values.resources | nindent 12 }}
+          {{- with .Values.extraVolumeMounts }}
+          volumeMounts:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
       priorityClassName: system-cluster-critical
+      {{- with .Values.extraVolumes }}
+      volumes:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
 {{- end }}
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -120,3 +120,19 @@ robot:
 podLabels: {}
 
 podAnnotations: {}
+
+# Mounts the specified volume to the hcloud-cloud-controller-manager container.
+extraVolumeMounts: []
+# # Example
+# extraVolumeMounts:
+#   - name: token-volume
+#     readOnly: true
+#     mountPath: /var/run/secrets/hcloud
+
+# Adds extra volumes to the pod.
+extraVolumes: []
+# # Example
+# extraVolumes:
+#   - name: token-volume
+#     secret:
+#       secretName: hcloud-token


apricote:
Awesome! Thanks @hegerdes.

Your values failed for me, the `env` is not properly defined and causes a YAML parser error. This works:

```yaml
env:
  HCLOUD_TOKEN_FILE:
    value: /var/run/secrets/hcloud/token
```

chart/templates/daemonset.yaml
```diff
@@ -92,5 +92,13 @@ spec:
             {{- end }}
           resources:
             {{- toYaml $.Values.resources | nindent 12 }}
+          {{- with .Values.extraVolumeMounts }}
+          extraVolumeMounts:
```
apricote, line 96:
```suggestion
          volumeMounts:
```

Same in chart/templates/deployment.yaml

[Pod API Spec](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#volumes-1)

chart/templates/daemonset.yaml
```diff
@@ -92,5 +92,13 @@ spec:
...
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
       priorityClassName: system-cluster-critical
+      {{- with .Values.extraVolumes }}
+      extraVolumes:
```
apricote, line 101:
```suggestion
      volumes:
```

Same in chart/templates/deployment.yaml

[Pod API Spec](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#volumes)

apricote (changes requested):

chart/templates/daemonset.yaml
hegerdes: Thanks!

hegerdes:
> Awesome! Thanks @hegerdes.
> 
> Your values failed for me, the `env` is not properly defined and causes a YAML parser error. This works:
> 
> ```yaml
> env:
>   HCLOUD_TOKEN_FILE:
>     value: /var/run/secrets/hcloud/token
> ```

Thanks, I updated the example.

I tested now with `helm upgrade --install ccm chart  --values chart/my-values.yaml`

apricote (approved):
Awesome! Looks good now. The e2e test failures are expected as we have no easy way to hand out API tokens to pull requests from external contributors.

PR #750, opened 2024-09-16
fix: change annotation LBSvcRedirectHTTP to "https"

### LINKED ISSUE #656: Inconsistency in the naming for load balancer annotation.
https://github.com/hetznercloud/hcloud-cloud-controller-manager/blob/e2b0ed63f6cd5a7499d356bacd784ded41df5f2a/internal/annotation/load_balancer.go#L161

Isn't this supposed to be http-redirect-https? Most guides also say that it's http-redirect-https, but in the code itself it's http-redirect-http.

a-mader:
Hi,

I think this should be renamed from "http-to-http" to "http-to-https" and this would be a breaking change.

Closes #656 

Kind regards

### DIFF
--- a/internal/annotation/load_balancer.go
+++ b/internal/annotation/load_balancer.go
@@ -162,7 +162,7 @@ const (
 	LBSvcHTTPManagedCertificateDomains Name = "load-balancer.hetzner.cloud/http-managed-certificate-domains"
 
 	// LBSvcRedirectHTTP create a redirect from HTTP to HTTPS. HTTPS only.
-	LBSvcRedirectHTTP Name = "load-balancer.hetzner.cloud/http-redirect-http"
+	LBSvcRedirectHTTP Name = "load-balancer.hetzner.cloud/http-redirect-https"
 
 	// LBSvcHTTPStickySessions enables the sticky sessions feature of Hetzner
 	// Cloud HTTP Load Balancers.


apricote:
Thanks for the PR @a-mader,

this is indeed a breaking change, so I will not merge it as is.

I wrote a bit about the origin of the current annotation and the alternative I would like to see in https://github.com/hetznercloud/hcloud-cloud-controller-manager/issues/656#issuecomment-2354629495.

If you want you can adjust this PR to match my suggestion in the linked comment.

a-mader:
Hi @apricote,

what do you think about having both annotations available, but marking the old one for deprecation?

e.g. 
- LBSvcRedirectHTTP marked for deprecation
- LBSvcRedirectHTTPS new annotation

```go
// MARKED FOR DEPRECATION (LBSvcRedirectHTTP create a redirect from HTTP to HTTPS. HTTPS only.)
LBSvcRedirectHTTP Name = "load-balancer.hetzner.cloud/http-redirect-http"

// LBSvcRedirectHTTP create a redirect from HTTP to HTTPS. HTTPS only. Bool either "true" or "false"
LBSvcRedirectHTTPS Name = "load-balancer.hetzner.cloud/http-redirect-https"
```

and update the documentation to use the new annotation.

Kind regards

apricote:
I would prefer to not mark `LBSvcRedirectHTTP` as deprecated. After all it is the one matching the API.

PR #752, opened 2024-09-28
Allow arbitrary length API token behind a flag

kamushadenes:
# Context
[We](https://github.com/altinity) have developed a soon-to-be-open-source proxy that forces specific labels in order to provide scoped API access, and that doesn't expose the real API token. This was created to have better control of resources inside the same project (as API tokens currently lack granularity), and to be able to use a single project securely, given that it isn't possible to create a project via the API.

One of it's operating modes is using JWT as a virtual self-validating token, which can't have a fixed size.

This support is required to make full use of it inside a Kubernetes cluster. 

The feature is behind a default-false flag so it shouldn't interfere with current behavior.

# Related
https://github.com/kubernetes/autoscaler/pull/7285
https://github.com/hetznercloud/csi-driver/pull/724

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -11,10 +11,11 @@ import (
 )
 
 const (
-	hcloudToken    = "HCLOUD_TOKEN"
-	hcloudEndpoint = "HCLOUD_ENDPOINT"
-	hcloudNetwork  = "HCLOUD_NETWORK"
-	hcloudDebug    = "HCLOUD_DEBUG"
+	hcloudToken                     = "HCLOUD_TOKEN"
+	hcloudAllowArbitraryLengthToken = "HCLOUD_ALLOW_ARBITRARY_LENGTH_TOKEN"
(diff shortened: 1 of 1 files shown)


apricote:
Hey @kamushadenes,

we talked about this today internally. We would prefer not to add any additional flags or environment variables to disable the check.

We also do not think that the check is strictly necessary. Instead we would prefer to change the errors when `len(token) != 64` to a warning message but allow the token. This would also help us in the future if we ever change the format of our API tokens.

Do you want to update your PRs to log warnings or should we do work on that?

kamushadenes:
Hey @apricote, thanks for getting back!

Makes total sense, I'll update the PRs in a couple hours when my day starts.

kamushadenes:
Done!

apricote (approved):

apricote (approved):
Was about to leave a comment regarding the unit test :D

Do not worry about the e2e tests, they do not work for PRs from forks.

PR #764, opened 2024-10-22
feat(instance): add label to distinguish servers from Cloud and Robot

lukasmetzner:
This update to the HCCM introduces a new label for Kubernetes nodes: `instance.hetzner.cloud/provided-by`. This label will be automatically applied to nodes, indicating their origin as either:
- `robot` for dedicated servers
- `cloud` for cloud-based virtual machines

With this features other components (e.g. csi-driver) can fine-tune their deployment requirements.

### DIFF
--- a/hcloud/instances.go
+++ b/hcloud/instances.go
@@ -33,6 +33,10 @@ import (
 	"github.com/hetznercloud/hcloud-go/v2/hcloud"
 )
 
+const (
+	ProvidedBy = "instance.hetzner.cloud/provided-by"
+)
+
 type instances struct {
 	client        *hcloud.Client
 	robotClient   robot.Client
@@ -273,6 +277,9 @@ func (s hcloudServer) Metadata(addressFamily config.AddressFamily, networkID int
 		NodeAddresses: hcloudNodeAddresses(addressFamily, networkID, s.Server),
 		Zone:          s.Datacenter.Name,
 		Region:        s.Datacenter.Location.Name,
+		AdditionalLabels: map[string]string{
+			ProvidedBy: "cloud",
+		},
 	}, nil
 }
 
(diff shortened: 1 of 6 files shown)


hcloud/instances.go
```diff
@@ -33,6 +33,10 @@ import (
 	"github.com/hetznercloud/hcloud-go/v2/hcloud"
 )
 
+const (
+	ProvidedBy = "instance.hetzner.cloud/provided-by"
```
jooola, line 37: Do we have a documentation for how to create a new label? 

According to https://kubernetes.io/docs/reference/labels-annotations-taints, we might want to use: `node.hetzner.cloud/provided-by` ?

Correct me if I am wrong, but since we manage the whole lifecycle of the nodes, shouldn't we use `node.hetzner.cloud/managed-by` ?

jooola:
The title of the PR is confusing, in our context, a server type might be for example `cx22`.

hcloud/instances.go
apricote: There is no documentation for defining new labels at the moment.

`instance.hetzner.cloud/` is the prefix used by the previous csi-driver label `instance.hetzner.cloud/is-root-server`, so the idea was to just stick with the same prefix.

> Correct me if I am wrong, but since we manage the whole lifecycle of the nodes, shouldn't we use node.hetzner.cloud/managed-by ?

We do not manage the whole lifecycle here. Creating nodes is out of scope for HCCM. Managed By is more the controller/application that creates the Kubernetes Object, so in this case it could be one of `kubelet` or `hcloud-cloud-controller-manager`.

---

Happy to hear other suggestions

apricote (commented):
To make the StorageClass selection work with `cluster-autoscaler`, the label also needs to be set in cluster-autoscaler: https://github.com/kubernetes/autoscaler/blob/878c982ae433b6424a65ad73caeaa37ed710c853/cluster-autoscaler/cloudprovider/hetzner/hetzner_node_group.go#L386-L392

apricote (commented):
Could you add the new label to these two locations?

- https://github.com/hetznercloud/hcloud-cloud-controller-manager/tree/main?tab=readme-ov-file#node-metadata-example
- https://github.com/hetznercloud/hcloud-cloud-controller-manager/blob/main/docs/robot.md#node-controller

apricote (approved):

PR #770, opened 2024-10-30
fix: invalid characters in label instance-type of robot servers

lukasmetzner:
Invalid characters like `™` should be removed when generating the instance-type label for Robot servers.

### DIFF
--- a/hcloud/instances_test.go
+++ b/hcloud/instances_test.go
@@ -380,7 +380,7 @@ func TestInstances_InstanceMetadataRobotServer(t *testing.T) {
 				ServerIP:      "233.252.0.123",
 				ServerIPv6Net: "2a01:f48:111:4221::",
 				ServerNumber:  321,
-				Product:       "robot-product 1",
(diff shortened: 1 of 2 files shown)


lukasmetzner:
[Syntax and character set](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set)
> The name segment is required and must be 63 characters or less, beginning and ending with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between.

There are some additional cases we do not cover yet like: 
- 63 chars or less
- alphanumeric and beginning or end

Should we do something specific about them, e.g trim label, trim characters and beginning or end? @apricote

apricote:
Looking at the list of previous models I do not think we run into any problems with length or non-alphanumeric characters for now: https://docs.hetzner.com/robot/dedicated-server/general-information/root-server-hardware#hardware-configuration

apricote (approved):

PR #771, opened 2024-10-30
ci: introduce helm snapshots and ci verification

lukasmetzner:
Introduces snapshots generated with helm template from default values, example production values and full values. A script to generate this snapshots is run in the CI with a check if they are up to date.


.pre-commit-config.yaml
```diff
@@ -14,13 +14,14 @@ repos:
 
       - id: check-json
       - id: check-yaml
-        exclude: ^(deploy/.*|chart/templates/.*)$
+        exclude: ^(deploy/.*|chart/templates/.*|chart/.snapshots/.*)$
```
apricote, line 17: I do think it makes sense to validate the generated snapshots for valid yaml. 

What was the problem that made you exclude the snapshots?

chart/example-prod.values.yaml
```diff
@@ -0,0 +1,14 @@
+replicaCount: 3
```
apricote, line 1: I dont think the prod files add any values here. The full template already does nearly everything we can do.

chart/.snapshots/full.values.yaml
```diff
@@ -0,0 +1,159 @@
...
+  # https://github.com/hetznercloud/hcloud-cloud-controller-manager/issues/395
+  route-reconciliation-period: 30s
+
+  # We do not use the webhooks feature and there is no need to bind a port that is unused.
+  # https://github.com/kubernetes/kubernetes/issues/120043
+  # https://github.com/hetznercloud/hcloud-cloud-controller-manager/issues/492
+  webhook-secure-port: "0"
+
+# Change deployment kind from "Deployment" to "DaemonSet"
+kind: Deployment
```
apricote, line 20: What do you think about changing this to `DaemonSet` so we have one snapshot for each?

lukasmetzner: Sounds good. Changed the type to DaemonSet for `full.values.yaml`.

apricote (approved):

apricote (changes requested):
CI fails because of the file rename.

apricote (approved):

PR #773, opened 2024-10-31
feat: emit event when robot server name and node name mismatch

lukasmetzner:
When the Robot server name is changed in the Robot console or via the API the node might be deleted by the node-lifecycle-manager. We now emit a warning event for this case.

### DIFF
--- a/hcloud/instances_util.go
+++ b/hcloud/instances_util.go
@@ -74,14 +74,14 @@ func getRobotServerByName(c robot.Client, node *corev1.Node) (server *hrobotmode
 	return server, nil
 }
 
-func getRobotServerByID(c robot.Client, id int, node *corev1.Node) (*hrobotmodels.Server, error) {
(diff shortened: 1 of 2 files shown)


apricote (approved):

hcloud/instances_util.go
```diff
@@ -92,6 +92,14 @@ func getRobotServerByID(c robot.Client, id int, node *corev1.Node) (*hrobotmodel
 
 	// check whether name matches - otherwise this server does not belong to the respective node anymore
 	if server.Name != node.Name {
+		i.recorder.Eventf(
+			node,
+			corev1.EventTypeWarning,
+			"PossibleNodeDeletion",
+			"Might be deleted by node-lifecycle-manager due to name mismatch; Node name \"%s\" differs from Robot name \"%s\"",
```
apricote, line 99: Tip: You can use `%q` to get a quoted string:

```suggestion
			"Might be deleted by node-lifecycle-manager due to name mismatch; Node name %q differs from Robot name %q",
```

PR #780, opened 2024-11-05
feat(load-balancer): ignore nodes that don't use known provider IDs

lukasmetzner:
With this feature the load balancer is more resilient to hybrid clusters as it skips nodes with unkown provider ids.

### DIFF
--- a/internal/hcops/load_balancer.go
+++ b/internal/hcops/load_balancer.go
@@ -613,7 +613,7 @@ func (l *LoadBalancerOps) ReconcileHCLBTargets(
 	for _, node := range nodes {
 		id, isCloudServer, err := providerid.ToServerID(node.Spec.ProviderID)
 		if err != nil {
-			if errors.Is(err, &providerid.UnkownPrefixError{}) {
+			if errors.As(err, new(*providerid.UnkownPrefixError)) {
 				// ProviderID has unknown prefix, cluster might have non-hccm nodes that can not be added to the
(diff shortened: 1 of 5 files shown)


internal/hcops/load_balancer.go
```diff
@@ -613,6 +613,18 @@ func (l *LoadBalancerOps) ReconcileHCLBTargets(
...
+				// Load Balancer. Emitting an event and ignoring that Node in this reconciliation loop.
+				l.Recorder.Eventf(
+					node,
+					corev1.EventTypeWarning,
+					"UnknownProviderIDPrefix",
+					"Node could not be added to Load Balancer for service %s because the provider ID does not match any known format",
+					svc.Name,
+				)
+				continue
+			}
```
apricote, line 627: Could you add a test for this in `TestLoadBalancerOps_ReconcileHCLBTargets`?

internal/providerid/providerid.go
```diff
@@ -24,6 +24,20 @@ const (
 	prefixRobotLegacy = "hcloud://bm-"
 )
 
+type UnkownPrefixError struct {
+	ProviderID string
+}
+
+func (e *UnkownPrefixError) Error() string {
+	return fmt.Sprintf(
+		"providerID does not have one of the the expected prefixes (%s, %s, %s): %s",
```
jooola, line 33:
```suggestion
		"Provider ID does not have one of the the expected prefixes (%s, %s, %s): %s",
```

Makes it consistent with the rest of the names.

PR #783, opened 2024-11-06
feat(load-balancer): Set IPMode to "Proxy" if load balancer is configured to use proxy protocol (#727)

lukasmetzner:
[KEP-1860](https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/1860-kube-proxy-IP-node-binding) introduced a new field `service.status.loadBalancer.ingress[].ipMode`: VIP | Proxy to indicate the behavior of the Load Balancer.

Currently users on IPVS-based networking setups can not access the Load Balancer IPs when they enable PROXY protocol.
Expected behavior

Users should always be able to access their services from inside of the cluster, even if they use IPVS and PROXY protocol.

IIUC we should set the IP Mode to Proxy for the IPs we return in the status.


lukasmetzner:
As Kubernetes version 1.28 is no longer supported we can skip the version check.

jooola (approved):

jooola:
> As Kubernetes version 1.28 is no longer supported we can skip the version check.

Seems like we still support it though, so we might need to drop support for 1.28 before releasing this?

PR #784, opened 2024-11-06
fix: request flood when misconfigured cluster-cidr causes dead routes to be kept

### LINKED ISSUE #482: fix: Request Flood when misconfigured cluster-cidr causes dead routes to be kept
## Scenario
- Cluster CIDR in hccm config does not correspond to real Pod CIDRs
- Routes are created for real Pod CIDRs
- Nodes are removed again by an autoscaler
- Routes are never deleted (since #432) because CIDR is outside the configured Cluster CIDRs
- With `ListRoutes`, the existing routes are always listed again, but the gateway IPs cannot be assigned to any servers because the servers are deleted.
- In case of a cache miss we reload the complete server list.
- Since several faulty routes exist without servers, we reload the server list several times per ListRoutes

## Effect

Depending on the amount of "dead routes", this can quickly exhaust the API rate limit by sending multiple `GET /servers` requests per second.

## Solutions

- Detect misconfiguration: In CreateRoute we can validate that the Route CIDR is contained within the configured Cluster CIDR
- Slower cache refresh: Refreshing the cache multiple times per iteration is not necessary, it should be enough to refresh the cache once per `ListRoutes` call

lukasmetzner:
Before adding a route, verify if the route CIDR is contained within the cluster CIDR. If it is not, return an error to prevent a flood of requests caused by persistent dead routes.

Fixes #482

### DIFF
--- a/hcloud/routes.go
+++ b/hcloud/routes.go
@@ -119,6 +119,18 @@ func (r *routes) CreateRoute(ctx context.Context, clusterName string, nameHint s
 		return fmt.Errorf("%s: %w", op, err)
 	}
 
+	hNetSize, _ := r.network.IPRange.Mask.Size()
+	destNetSize, _ := cidr.Mask.Size()
+
+	if !(r.network.IPRange.Contains(cidr.IP) && destNetSize >= hNetSize) {
+		return fmt.Errorf(
+			"route CIDR %s is not contained within CIDR %s of hcloud network %d",
+			route.DestinationCIDR,
+			r.network.IPRange.String(),
+			r.network.ID,
+		)
+	}
+
 	doesRouteAlreadyExist, err := r.checkIfRouteAlreadyExists(ctx, route)
 	if err != nil {
 		return fmt.Errorf("%s: %w", op, err)
--- a/hcloud/routes_test.go
+++ b/hcloud/routes_test.go
@@ -83,6 +83,15 @@ func TestRoutes_CreateRoute(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Unexpected error: %v", err)
 	}
+
+	err = routes.CreateRoute(context.TODO(), "my-cluster", "routeFail", &cloudprovider.Route{
+		Name:            "route",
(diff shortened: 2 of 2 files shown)


hcloud/routes.go
```diff
@@ -119,6 +119,18 @@ func (r *routes) CreateRoute(ctx context.Context, clusterName string, nameHint s
...
+	hNetSize, _ := r.network.IPRange.Mask.Size()
+	destNetSize, _ := cidr.Mask.Size()
+
+	if !(r.network.IPRange.Contains(cidr.IP) && destNetSize >= hNetSize) {
+		return fmt.Errorf(
+			"route CIDR %s is not contained within CIDR %s of hcloud network %d",
+			route.DestinationCIDR,
+			r.network.IPRange.String(),
+			r.network.ID,
+		)
```
apricote, line 131: I think we should also add the Node Name, on the other hand the network id is irrelevant for this

```suggestion
		return fmt.Errorf(
			"route CIDR %s is not contained within CIDR %s of hcloud network %d",
			route.DestinationCIDR,
			r.network.IPRange.String(),
			r.network.ID,
		)
```

hcloud/routes_test.go
```diff
@@ -83,6 +83,15 @@ func TestRoutes_CreateRoute(t *testing.T) {
...
 	}
+
+	err = routes.CreateRoute(context.TODO(), "my-cluster", "routeFail", &cloudprovider.Route{
+		Name:            "route",
+		TargetNode:      "node15",
+		DestinationCIDR: "172.16.0.0/16",
+	})
+	if err == nil {
+		t.Fatalf("Expected an error because DestinationCIDR is not within the cluster CIDR, but received nil instead")
+	}
```
jooola, line 94: Could you test the error string too, please? Maybe with `require.ErrorRegexp` or `require.Error` and `require.Regexp`

lukasmetzner:
Already fixed in #675. 
Emitting a warning event and log message will be done in #793.

PR #793, opened 2024-11-18
feat(routes): emit warning if cluster cidr is misconfigured

lukasmetzner:
Before adding a route, verify if the route CIDR is contained within the configured cluster CIDR. If it is not, emit a warning event and log message.

### DIFF
--- a/hcloud/routes.go
+++ b/hcloud/routes.go
@@ -119,6 +119,18 @@ func (r *routes) CreateRoute(ctx context.Context, clusterName string, nameHint s
 		return fmt.Errorf("%s: %w", op, err)
 	}
 
+	hNetSize, _ := r.network.IPRange.Mask.Size()
+	destNetSize, _ := cidr.Mask.Size()
+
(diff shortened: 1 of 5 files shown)


lukasmetzner:
@apricote 

> Could you split up this PR into one that refactors the init code and the second one that emits the warning?

Created #794 for the refactoring. Afterward, we could rebase from main and merge this one, as we already referenced this PR in some answers.

hcloud/routes.go
```diff
@@ -17,6 +20,8 @@ import (
 	"github.com/hetznercloud/hcloud-go/v2/hcloud"
 )
 
+const DefaultClusterCIDR = "10.244.0.0/16"
```
apricote, line 23: Could you move this to `routes_test.go`? This is not a code-default, so I would prefer to have it in the tests to avoid any confusion.

apricote (changes requested):
One small change + rebase

apricote (approved):

PR #828, opened 2024-12-17
feat(load-balancer): emit warning if unsupported port protocol is configured

lukasmetzner:
If an unsupported port protocol is configured a warning event and log message is emitted.

### DIFF
--- a/internal/hcops/load_balancer.go
+++ b/internal/hcops/load_balancer.go
@@ -845,17 +845,6 @@ func (l *LoadBalancerOps) ReconcileHCLBServices(
 		return false, fmt.Errorf("%s: %v", op, err)
 	}
 
-	for _, port := range svc.Spec.Ports {
-		if port.Protocol != "" && port.Protocol != corev1.ProtocolTCP {
-			err := fmt.Errorf(
-				"configured unsupported Hetzner Cloud load balancer protocol %s for service with name %s",
-				port.Protocol,
-				svc.Name,
-			)
-			return false, err
-		}
-	}
-
 	hclbListenPorts := make(map[int]bool, len(lb.Services))
 	for _, hclbService := range lb.Services {
 		hclbListenPorts[hclbService.ListenPort] = true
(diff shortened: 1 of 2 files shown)


jooola:
If I understand this, we only trigger a warning event, and we don't fail?

lukasmetzner:
> If I understand this, we only trigger a warning event, and we don't fail?

We do fail, the load balancer is not reconciled and we return an error to the cloud-provider call site, which will dispatch an event and log message.

internal/hcops/load_balancer.go
```diff
@@ -845,6 +845,17 @@ func (l *LoadBalancerOps) ReconcileHCLBServices(
...
 
+	for _, port := range svc.Spec.Ports {
+		if port.Protocol != "" && port.Protocol != corev1.ProtocolTCP {
+			err := fmt.Errorf(
+				"configured unsupported Hetzner Cloud load balancer protocol %s for service with name %s",
+				port.Protocol,
+				svc.Name,
+			)
+			return false, err
+		}
```
jooola, line 856: If we don't plan to fail, shouldn't we continue processing the function, instead of returning? Feels like a breaking change.

jooola: I always thought we didn't want to break user land for this issue. So I expect this to apply as much as possible and only trigger warning events where we can't/don't support the feature.

But maybe you already discussed a different plan with @apricote?

lukasmetzner: No you are totally right. It would be a breaking change. I missed that and just thought about how I would like to have it :sweat_smile:

Will adapt the code accordingly and dispatch an event/log msg here instead of failing.

apricote: Yea, I would also prefer it if we could fail completely, but that would be a breaking change.

internal/hcops/load_balancer.go
```diff
@@ -877,6 +866,21 @@ func (l *LoadBalancerOps) ReconcileHCLBServices(
...
+            )
+			l.Recorder.Event(
+				svc,
+				corev1.EventTypeWarning,
+				"UnsupportedProtocolConfigured",
+                warnMsg,
+			)
+			klog.Warning(warnMsg)
+		}
+
```
jooola: Again those weird indents, is it GitHub that is messing things up or are you using space instead of tabs?

jooola: Aah sorry, outdated changes.

lukasmetzner: I switched to nvim recently and pretty sure I have something misconfigured there. Need to fix this. But golangci-lint fixes it, so good that we added it to pre-commit ;)

jooola (approved):
Could you please update the PR title and description to reflect the new changes?

Ah and the unit tests :)

lukasmetzner:
> Could you please update the PR title and description to reflect the new changes?

Yes. Need to update the unit tests before that.

PR #829, opened 2024-12-17
ci: added golangci-lint to pre-commit

lukasmetzner:
Added golangci-lint run hook to pre-commit configuration and removed separate CI step.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,22 +36,6 @@ jobs:
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
 
-  lint:
-    name: Lint
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-go@v5
-        with:
-          go-version-file: go.mod
-
-      - uses: golangci/golangci-lint-action@v6
-        with:
-          version: v1.62.2 # renovate: datasource=github-releases depName=golangci/golangci-lint
-
-          # In general linting is quite fast with warm caches, but a fresh run might take some time.
-          args: --timeout 5m
-
   deploy-manifests:
     name: Check deployment manifests
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
       - uses: actions/cache@v4
         with:
           path: ~/.cache/golangci-lint
-          key: pre-commit-3|${{ env.pythonLocation }}|${{ hashFiles('.pre-commit-config.yaml') }}
+          key: pre-commit-golangci-lint|${{ env.pythonLocation }}|${{ hashFiles('.pre-commit-config.yaml') }}
 
       - uses: pre-commit/action@v3.0.1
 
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -46,6 +46,12 @@ repos:
         entry: shfmt -i 2 -ci -sr -kp -w
         types: [shell]
 
+  - repo: https://github.com/golangci/golangci-lint/
+    rev: v1.62.2
+    hooks:
+      - id: golangci-lint
```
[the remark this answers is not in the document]
lukasmetzner: Good call, as we use it in the CI.

jooola (approved):
I would just check how long the cold run takes and see if the cache is persisted across ci runs, if not we might need to investigate why the cache is not preserved.

Other than that looks good from my side, maybe @apricote is against this, I'll leave the final decision to him.

.pre-commit-config.yaml
```diff
@@ -46,6 +46,13 @@ repos:
         entry: shfmt -i 2 -ci -sr -kp -w
         types: [shell]
 
+  - repo: https://github.com/golangci/golangci-lint
+    rev: v1.62.2
+    hooks:
+      - id: golangci-lint-full
+        args: [--timeout=5m]
+        language_version: 1.23.4 # renovate: datasource=golang-version
+
```
jooola, line 55: The shell check and the shfmt were next to each because they both focus on bash script, I would put this hook either before or after.

.github/workflows/ci.yml
```diff
@@ -16,6 +16,11 @@ jobs:
         with:
           python-version: 3.x
 
+      - uses: actions/cache@v4
+        with:
+          path: ~/.cache/golangci-lint
+          key: pre-commit-3|${{ env.pythonLocation }}|${{ hashFiles('.pre-commit-config.yaml') }}
```
jooola, line 22: Might not be the best idea to use the same cache key as pre-comit. Maybe you can do:


`pre-commit-golangci-lint|${{ env.pythonLocation }}|${{ hashFiles('.pre-commit-config.yaml') }}`

lukasmetzner: According to the docs a cache can handle multiple directories [1]. Not sure though if using the cache action twice with different dirs on the same cache key gets us the same behavior. Would be interesting to test.

Our pre-commit-3 cache takes 170MB, while the pre-commit-golangci-lint cache has a size of 1.7MB. Might be more efficient to do a single cache restore with both dirs(?). Apparently its backbone is Azure blob storage [2].

I think we can leave it at using two caches, just found the topic interesting.

[1] https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#input-parameters-for-the-cache-action
[2] https://depot.dev/blog/github-actions-cache

lukasmetzner: I think we might want to have a different cache key. When using `pre-commit-golangci-lint|${{ env.pythonLocation }}|${{ hashFiles('.pre-commit-config.yaml') }}` the cache will never be updated until we either edit the pre-commit config or a new Python env is setup through the Python action.

We probably should include some hashes of our sources.

jooola: If you want to be really precise, you might need to play with the `restore-key` action input, and use the input key you specified.

Issue #790, opened 2024-11-12
Robot servers cannot be initialised


vladimirtiukhtin:
### TL;DR

```
E1112 22:16:14.736364       1 node_controller.go:240] error syncing 'XXX': failed to get node modifiers from cloud provider: provided node ip for node "XXX" is not valid: failed to get node address from cloud provider that matches ip: 10.103.1.103, requeuing
```

### Expected behavior

Robot servers with no taints

### Observed behavior

As described above

### Additional information

```
        env:
        - name: NODE_NAME
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: spec.nodeName
        - name: HCLOUD_LOAD_BALANCERS_LOCATION
          value: hel1
        - name: HCLOUD_LOAD_BALANCERS_USE_PRIVATE_IP
          value: "true"
        - name: HCLOUD_LOAD_BALANCERS_DISABLE_PRIVATE_INGRESS
          value: "true"
        - name: HCLOUD_TOKEN
          valueFrom:
            secretKeyRef:
              key: token
              name: hcloud
        - name: HCLOUD_NETWORK
          valueFrom:
            secretKeyRef:
              key: network
              name: hcloud
        - name: HCLOUD_NETWORK_ROUTES_ENABLED
          value: "false"
        - name: ROBOT_ENABLED
          value: "true"
        - name: ROBOT_USER
          valueFrom:
            secretKeyRef:
              key: robot_username
              name: hcloud
        - name: ROBOT_PASSWORD
          valueFrom:
            secretKeyRef:
              key: robot_password
              name: hcloud
```

lukasmetzner:
Hey,

does your host name of the Robot server match with the name as shown in the Robot console?

Best Regards,
Lukas

lukasmetzner:
Hey,

We are happy to announce that we improved our Robot support in HCCM version [v1.24.0](https://github.com/hetznercloud/hcloud-cloud-controller-manager/releases/tag/v1.24.0). The HCCM now forwards `InternalIPs` by default on Robot nodes when the `--node-ip` flag is used. If the provided IP is not already registered as an `ExternalIP` and matches the expected address family, it will be forwarded automatically during initialization.

This allows the use of vSwitch IPs in private networks and Load Balancers.

This does not change anything with respect to the restrictions of the routing capability. This is still not supported by the underlying product.

Best Regards,
Lukas

Issue #760, opened 2024-10-20
Robot Servers InternalIP is not set in node manifest


mehrdadpfg:
### TL;DR

Dedicated Servers InternalIP should be set in the node manifests like the cloud servers. It is an important bug since metrics-server can not reach the servers and also the kubelet has been configured by the InternalIP.( --node flag )

### Expected behavior

```
status:
  addresses:
  - address: x.x.x.x
    type: InternalIP
  - address: k8s-node
    type: Hostname
  - address: x.x.x.x.
    type: ExternalIP
 ```
 expected to see the internalIP in the dedicated servers as well

### Observed behavior

```
status:
  addresses:
  - address: k8s-node
    type: Hostname
  - address: x.x.x.x.
    type: ExternalIP
 ```

### Log output
constantly trying to find the dedicated server privateIP in cloud
```plain text
 19:14:35.804948       1 node_controller.go:389] Failed to update node addresses for node "k8s-node": failed to get node address from cloud provider that matches ip: x.x.x.x
```

mehrdadpfg:
can anybody help us here?

pat-s:
#696 seems to address this though the PR seems to be somewhat stuck.

lukasmetzner:
Hey,

We are happy to announce that we improved our Robot support in HCCM version [v1.24.0](https://github.com/hetznercloud/hcloud-cloud-controller-manager/releases/tag/v1.24.0). The HCCM now forwards `InternalIPs` by default on Robot nodes when the `--node-ip` flag is used. If the provided IP is not already registered as an `ExternalIP` and matches the expected address family, it will be forwarded automatically during initialization.

This allows the use of vSwitch IPs in private networks and Load Balancers.

Best Regards,
Lukas

Issue #708, opened 2024-07-31
HCCM Controller don't set metadata to the Nodes at all (nodes are connected, using internal network)


DmitrySkrotski:
### TL;DR

I set up cluster on hcloud (controlplanes are cloud, workers are robots + cloud) and I am expecting that the HCCM will populate node objects with metadata.

### Expected behavior

HCCM will populate  information about the server to the Node object as described here [hccm](https://github.com/hetznercloud/hcloud-cloud-controller-manager/blob/main/docs/robot.md#node-controller)

### Observed behavior

Nodes joined using join.yml file
```apiVersion: kubeadm.k8s.io/v1beta3
kind: JoinConfiguration
discovery:
  bootstrapToken:
    apiServerEndpoint: 10.195.4.250:6443
    caCertHashes:
    - sha256:xxxxxxxxx
    token: xxxxxxx.xxxxxxxxxxxxxx
nodeRegistration:
  name: prod-k8s-metall01
  criSocket: unix:///var/run/containerd/containerd.sock
  kubeletExtraArgs:
    node-ip: 10.195.10.2
```

HCCM is setup using helm 
```
helm install hcloud-controller-manager hcloud/hcloud-cloud-controller-manager --set robot.enabled=true --set netwo
rk.enabled=false -n kube-system
```

secret hcloud persists
![image](https://github.com/user-attachments/assets/ccc794fc-7caf-44fa-9367-35f521ef2e1b)
HCCM logs:
![image](https://github.com/user-attachments/assets/10a497a5-6ddd-4d7a-9d7a-54d5c4267cb5)
pod initialization log:

### Log output

```
plain text
I0731 08:29:07.120000       1 controllermanager.go:169] Version: v0.0.0-master+$Format:%H$
I0731 08:29:07.139351       1 secure_serving.go:213] Serving securely on [::]:10258
I0731 08:29:07.139953       1 tlsconfig.go:240] "Starting DynamicServingCertificateController"
I0731 08:29:07.140307       1 requestheader_controller.go:169] Starting RequestHeaderAuthRequestController
I0731 08:29:07.140913       1 shared_informer.go:313] Waiting for caches to sync for RequestHeaderAuthRequestController
I0731 08:29:07.140361       1 configmap_cafile_content.go:202] "Starting controller" name="client-ca::kube-system::extension-apiserver-authentication::client-ca-file"
[...]

Issue #703, opened 2024-07-22
HCCM should work with FQDN Servernames and Private IPs as Node IPs


tehnic-take3:
### TL;DR

Hello,

first - i prefer to use FQDN servernames like "k8s-control-plane1.mydomain.com", but then i get error messages in the HCCM log like:
```
I0722 xx:xx:xx.xxxxxx       1 node_controller.go:425] Initializing node k8s-control-plane1 with cloud provider
I0722 xx:xx:xx.xxxxxx       1 node_controller.go:229] error syncing 'k8s-control-plane1': failed to get instance metadata for node k8s-control-plane1: hcloud/instancesv2.InstanceMetadata: failed to get instance metadata: no matching server found for node 'k8s-control-plane1': server not found, requeuing
E0722 xx:xx:xx.xxxxxx       1 node_controller.go:240] error syncing 'k8s-control-plane1': failed to get instance metadata for node k8s-control-plane1: hcloud/instancesv2.InstanceMetadata: failed to get instance metadata: no matching server found for node 'k8s-control-plane1': server not found, requeuing
```

I suggest to have a HCCM configuration item to use the FQDN of all nodes for cloud provider API.

Second - after changing servername to one without domain, i get error messages like:
```
I0722 xx:xx:xx.xxxxxx       1 node_controller.go:425] Initializing node k8s-control-plane1 with cloud provider
I0722 xx:xx:xx.xxxxxx       1 node_controller.go:229] error syncing 'k8s-control-plane1': failed to get node modifiers from cloud provider: provided node ip for node "k8s-control-plane1" is not valid: failed to get node address from cloud provider that matches ip: 10.x.x.x, requeuing
E0722 xx:xx:xx.xxxxxx       1 node_controller.go:240] error syncing 'k8s-control-plane1': failed to get node modifiers from cloud provider: provided node ip for node "k8s-control-plane1" is not valid: failed to get node address from cloud provider that matches ip: 10.x.x.x, requeuing
```

I  used "KUBELET_EXTRA_ARGS=--cloud-provider=external --node-ip=10.x.x.x" for kubelet configuration on control-plane (KUBELET_EXTRA_ARGS=--node-ip=10.x.x.x on workers) where 10.x.x.x is the private IP of the node.

Currently i use antrea as CNI with geneve link between the nodes (using the private node IP addresses).

[...]

PR #851, opened 2025-01-23
feat: Allow to treat kubelet IPs as internal IPs when using robot nodes

### LINKED ISSUE #676: Local IPs with Robots server on vSwitch not working
### TL;DR

Node not adding if kubelet IP is set to internal network via Vswitch

### Expected behavior

Node succesfully updated with Node controller

### Observed behavior

Node broken and stays as uninitialized 
### Minimal working example

_No response_

### Log output

```plain text
E0702 15:17:50.125761       1 node_controller.go:240] error syncing 'kn2-stage.htzn': failed to get node modifiers from cloud provider: provided node ip for node "kn2-stage.htzn" is not valid: failed to get node address from cloud provider that matches ip: 192.168.0.131, requeuing
I0702 15:17:50.135997       1 node_controller.go:431] Initializing node kn2-stage.htzn with cloud provider
```

### Additional information

NAME             STATUS     ROLES           AGE    VERSION   INTERNAL-IP     EXTERNAL-IP   OS-IMAGE         KERNEL-VERSION   CONTAINER-RUNTIME
km1-stage        Ready      control-plane   14d    v1.27.2   192.168.0.3     <none>        Talos (v1.7.0)   6.6.28-talos     containerd://1.7.15
km2-stage        Ready      control-plane   14d    v1.27.2   192.168.0.5     <none>        Talos (v1.7.0)   6.6.28-talos     containerd://1.7.15
km3-stage        Ready      control-plane   18d    v1.27.2   192.168.0.4     <none>        Talos (v1.7.0)   6.6.28-talos     containerd://1.7.15
kn1-stage.htzn   Ready      <none>          155m   v1.27.2   192.168.0.130   <none>        Talos (v1.7.4)   6.6.32-talos     containerd://1.7.16
kn2-stage        Ready      egress-proxy    13d    v1.27.2   192.168.0.131   <none>        Talos (v1.7.4)   6.6.32-talos     containerd://1.7.16

codablock:
This introduces a new environment variable
(ROBOT_PROVIDED_NODE_IP_INTERNAL) which causes HCC to treat IPs give via
kubelet's --node-ip flag as internal node IP. The assumption here is
that robot nodes will always treat IP reported via the robot webservice
as external, because robot nodes don't know the concept of private IPs
directly. The only way to introduce private IPs is via other means, e.g.
vSwitches or VPNs. In that case, operators would usually provide the
internal IP to kubelet (via --node-ip), so that the kubelet only listens
on the internal IP instead of the public IP.

Kubelet provides this IP to the k8s Node object via the
"alpha.kubernetes.io/provided-node-ip" annotation, but only when the
cloud provider is set to external (which is of course the case for us).
We can leverage this annotation in HCC and use it to fill the
status.addresses field with this IP (in addition to the already added
NodeExternalIP and NodeHostnName).

This commit/PR also changes the load balancer provider to pass this same
IP to the target configuration in case ROBOT_PROVIDED_NODE_IP_INTERNAL
is enabled and "load-balancer.hetzner.cloud/use-private-ip=true" is
passed to the LoadBalancer SVC.

With these changes a robot node can go fully private and close down all
public ports via the robot firewall or other means. It essentially can
be treated like a private cloud node.

Fixes #676 #790 #760 #708 #635 #703 #571 #620
Replaces #696 (as it doesn't require static configuration of IPs at controller installation time)


codablock:
@lukasmetzner Cool, happy to switch over to the official solution for this problem :)

For me, and I assume for some/many others as well, the important part is that there is no upfront configuration/mapping of node <-> ips required, I assume that your solution will cover this?

lukasmetzner:
> For me, and I assume for some/many others as well, the important part is that there is no upfront configuration/mapping of node <-> ips required, I assume that your solution will cover this?

There won't be any additional mappings that need to be configured. We will also work with the value the user provides via `--node-ip` to the kubelet.

lukasmetzner:
Hey,

we have implemented your suggested methods in release v1.24.0. Configuring custom InternalIPs via the kubelet flag `--node-ip` is now possible.

Thank you for your pull request and the suggested solutions ^^

I will go ahead and close this PR.

Best Regards
Lukas

PR #865, opened 2025-02-10
feat(robot): forward InternalIPs by default on Robot nodes

lukasmetzner:
Whenever a user configures the `--node-ip` flag, we forward the configured IPs. We validate if the IP already exists as an ExternalIP and does not collide with the configured address family. This improves the Robot support for HCCM, as users can configure private IPs for Robot nodes.

Forwarding only happens during the initialization phase of HCCM and should therefore not break existing clusters.

This feature can be disabled by setting the environment variable `ROBOT_FORWARD_INTERNAL_IPS` to `false`.

Co-authored-by: Alexander Block <ablock84@gmail.com>


hcloud/instances.go
```diff
@@ -283,13 +284,22 @@ func robotNodeAddresses(
 		)
 	}
 
+	if cfg.Robot.ForwardInternalIPs {
+		for _, addr := range node.Status.Addresses {
+			if addr.Type == corev1.NodeInternalIP {
+				addresses = append(addresses, addr)
+			}
+		}
+	}
```
apricote, line 293: How does this behave in case a user configured `--node-ip=<external-ipv4>`? Is that IP then added twice on the node object?

internal/config/config.go
```diff
@@ -18,11 +18,12 @@ const (
...
-	robotUser              = "ROBOT_USER"
-	robotPassword          = "ROBOT_PASSWORD"
-	robotCacheTimeout      = "ROBOT_CACHE_TIMEOUT"
-	robotRateLimitWaitTime = "ROBOT_RATE_LIMIT_WAIT_TIME"
+	robotEnabled                   = "ROBOT_ENABLED"
+	robotUser                      = "ROBOT_USER"
+	robotPassword                  = "ROBOT_PASSWORD"
+	robotCacheTimeout              = "ROBOT_CACHE_TIMEOUT"
+	robotRateLimitWaitTime         = "ROBOT_RATE_LIMIT_WAIT_TIME"
+	robotDisableForwardInternalIPs = "ROBOT_DISABLE_FORWARD_INTERNAL_IPS"
```
apricote, line 26: What do you think about using `ROBOT_FORWARD_INTERNAL_IPS` with a default of `true`? Double negation is always prone for errors.

hcloud/instances.go
lukasmetzner: It is added twice with the respective types `InternalIP` and `ExternalIP` in the `status.addresses` field.

apricote: Not sure if that might cause problems? We could filter for it here, and not add it as a `NodeInternalIP` if its already under `NodeExternalIP`. That matches the current behaviour most.

internal/config/config.go
apricote: > So let's switch to ROBOT_FORWARD_INTERNAL_IPS?

I am happy with that. Eventually I think we should add something more sane than 100 env vars for configuration anyway. (Flags, Config File)

hcloud/instances.go
lukasmetzner: I can't think of any issues that could arise. Its probably more sane to filter here. Shall we emit a warning event+log, or info log?

apricote: Event+Log sounds good, will also not spam the user because it only happens once during the adoption.

hcloud/instances.go
```diff
@@ -318,6 +356,7 @@ func (s hcloudServer) Metadata(networkID int64, cfg config.HCCMConfiguration) (*
 type robotServer struct {
 	*hrobotmodels.Server
 	robotClient robot.Client
+	recoder     record.EventRecorder
```
apricote, line 360:
```suggestion
	recorder     record.EventRecorder
```

Also needs to be fixed anywhere this field is used.

hcloud/instances_test.go
```diff
@@ -644,6 +645,25 @@ func TestNodeAddressesRobotServer(t *testing.T) {
 				{Type: corev1.NodeExternalIP, Address: "203.0.113.7"},
 			},
 		},
+		{
+			name:          "public ipv4 and internal ip",
```
apricote, line 649: Could you add test cases for the conditions where we do not add the IPs? ie `MisconfiguredInternalIP`. Its enough for me to validate that they are not in the returned IPs, no need to check that the event is emitted.

apricote (changes requested):

apricote (approved):

PR #884, opened 2025-03-06
ci: fix pre-commit golang version

lukasmetzner:
As there was no `depName` defined with the `golang-version` `datasource`, our custom Renovate regex manager did not detect this as a dependency. We can also use the git tags from the official repository as [`golang-version`](https://docs.renovatebot.com/modules/datasource/golang-version/) is using this as a source AFAIK.

Question is, if we really need this check at all. Pre-commit will use the newest Golang version by default, which should be sufficient as `golangci-lint` will use the go version specified in the `go.mod`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -56,4 +56,3 @@ repos:
     hooks:
       - id: golangci-lint-full
         args: [--timeout=5m]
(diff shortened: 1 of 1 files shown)


jooola:
I think we could drop the language version field, and reconsider it if it breaks something. But so far we never had any issues.

jooola (approved):

PR #885, opened 2025-03-11
ci: fix golangci-lint using go toolchain version in pre-commit

lukasmetzner:
Fixes pre-commit CI pipeline of #881.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,15 +12,6 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      # Pull the toolchain version from go.mod
-      # setup-go feature request: https://github.com/actions/setup-go/issues/457
-      - id: toolchain-version
-        run: echo "version=$(sed -ne '/^toolchain /s/^toolchain go//p' go.mod)" >> "$GITHUB_OUTPUT"
-
-      - uses: actions/setup-go@v5
-        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -12,6 +12,15 @@ jobs:
...
 
+      # Pull the toolchain version from go.mod
+      # setup-go feature request: https://github.com/actions/setup-go/issues/457
+      - id: toolchain-version
+        run: echo "version=$(sed -ne '/^toolchain /s/^toolchain go//p' go.mod)" >> "$GITHUB_OUTPUT"
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version: ${{ steps.toolchain-version.outputs.version }}
+
```
jooola, line 23: Might be worth adding the language_version property in the pre-commit hook. Also we might want to upstream this change, as upstream is broken if we don't use the right golang version.

lukasmetzner: https://github.com/actions/setup-go/pull/460

Seems like there is an already existing implementation, which unfortunately does not get much attention. Not sure if it is worth the effort to contribute upstream.

jooola (approved):
I assume the version is the one before the jump to go1.24.

lukasmetzner:
> I assume the version is the one before the jump to go1.24.

It corresponds to the current toolchain version in main.

PR #886, opened 2025-03-11
ci: group golang deps in renovate

lukasmetzner:
Co-authored-by: jonas.lammler@hetzner-cloud.de

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -3,5 +3,18 @@
   "extends": [
     "github>hetznercloud/.github//renovate/default",
     "github>hetznercloud/.github//renovate/golang"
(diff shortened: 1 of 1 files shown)


apricote (approved):
Should we just enable this in our default config file so it applies to all go projects? https://github.com/hetznercloud/.github/blob/main/renovate/golang.json

jooola:
I am not sure if this is relevant in all our golang projects, we should go over our project and check first and then move it to the default config.

I'd like to try it out here first. But I'll take the idea with me.

jooola (approved):

PR #898, opened 2025-03-27
feat(load-balancer): enable use-private-ip annotation for Robot servers

lukasmetzner:
This feature enables the annotation `load-balancer.hetzner.cloud/use-private-ip` for clusters that include Robot servers.

We recently introduced forwarding configured InternalIPs of Robot nodes (#865). With this addition we can use these IPs and configure them as load balancer targets, when `load-balancer.hetzner.cloud/use-private-ip` is set.

### DIFF
--- a/docs/how-to-robot-vswitch-load-balancer.md
+++ b/docs/how-to-robot-vswitch-load-balancer.md
@@ -6,7 +6,9 @@ As a result, the annotation `load-balancer.hetzner.cloud/use-private-ip` can be 
 
 ## Configuration
 
-To configure this, enable Robot support as outlined in the [robot setup guide](./robot.md). Since the HCCM needs to fetch network data, provide the network ID using the `HCLOUD_NETWORK` environment variable. To prevent the HCCM from making other network changes, disable networking in the Helm chart and set `HCLOUD_NETWORK_ROUTES_ENABLED=false` to turn off the routes controller. Use the following snippet as a reference.
+To configure this, enable Robot support as outlined in the [robot setup guide](./robot.md). Since the HCCM needs to fetch network data, provide the network ID using the `HCLOUD_NETWORK` environment variable.
+
+To disable the Routes controller, which is incompatible with vSwitches, disable networking in the Helm chart and set `HCLOUD_NETWORK_ROUTES_ENABLED=false`. Use the following snippet as a reference.
 
 ```yaml
 networking:
--- a/docs/how-to-robot-vswitch-load-balancer.md
+++ b/docs/how-to-robot-vswitch-load-balancer.md
@@ -1,6 +1,6 @@
 # How to attach load balancers to Robot private IPs
 
-With our HCCM release v1.24.0 we introduced the option to configure InternalIPs for Robot servers. This allows creating a cluster with private networks and a mixture of Robot and Cloud servers. Using the routing feature of private networks is not supported, so it requires a CNI plugin with encapsulation methods, such as Cilium with routing mode `tunnel`. Load balancers can have targets of type IP, which can either be a public or private (vSwitch) IP of a Robot server ([API reference](https://docs.hetzner.cloud/#load-balancer-actions-add-target)).
+With the v1.24.0 release we introduced the option to configure Internal IPs for Robot servers. This allows creating a cluster with private networks and a mixture of Robot and Cloud servers. Using the routing feature of private networks is not supported, so this requires a CNI plugin with encapsulation methods, such as Cilium with routing mode `tunnel`. Load Balancers can have targets of type IP, which can either be a public or private (vSwitch) IP of a Robot server ([API reference](https://docs.hetzner.cloud/#load-balancer-actions-add-target)).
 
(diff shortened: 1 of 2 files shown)


docs/how-to-robot-vswitch-load-balancer.md
```diff
@@ -0,0 +1,44 @@
+# How to attach load balancers to Robot private IPs
+
+With our HCCM release v1.24.0 we introduced the option to configure InternalIPs for Robot servers. This allows creating a cluster with private networks and a mixture of Robot and Cloud servers. Using the routing feature of private networks is not supported, so it requires a CNI plugin with encapsulation methods, such as Cilium with routing mode `tunnel`. Load balancers can have targets of type IP, which can either be a public or private (vSwitch) IP of a Robot server ([API reference](https://docs.hetzner.cloud/#load-balancer-actions-add-target)).
```
apricote, line 3:
```suggestion
With the v1.24.0 release we introduced the option to configure Internal IPs for Robot servers. This allows creating a cluster with private networks and a mixture of Robot and Cloud servers. Using the routing feature of private networks is not supported, so this requires a CNI plugin with encapsulation methods, such as Cilium with routing mode `tunnel`. Load Balancers can have targets of type IP, which can either be a public or private (vSwitch) IP of a Robot server ([API reference](https://docs.hetzner.cloud/#load-balancer-actions-add-target)).
```

docs/how-to-robot-vswitch-load-balancer.md
```diff
@@ -0,0 +1,44 @@
...
+
+As a result, the annotation `load-balancer.hetzner.cloud/use-private-ip` can be set, if the Robot server is connected to a private network and its IP is of type [InternalIP](https://kubernetes.io/docs/reference/node/node-status/#addresses).
+
+## Configuration
+
+To configure this, enable Robot support as outlined in the [robot setup guide](./robot.md). Since the HCCM needs to fetch network data, provide the network ID using the `HCLOUD_NETWORK` environment variable. To prevent the HCCM from making other network changes, disable networking in the Helm chart and set `HCLOUD_NETWORK_ROUTES_ENABLED=false` to turn off the routes controller. Use the following snippet as a reference.
```
apricote, line 9:
```suggestion
To configure this, enable Robot support as outlined in the [robot setup guide](./robot.md). Since the HCCM needs to fetch network data, provide the network ID using the `HCLOUD_NETWORK` environment variable.

To disable the Routes controller, which is incompatible with vSwitches, disable networking in the Helm chart and set `HCLOUD_NETWORK_ROUTES_ENABLED=false`. Use the following snippet as a reference.
```

internal/hcops/load_balancer.go
```diff
@@ -645,6 +645,27 @@ func (l *LoadBalancerOps) ReconcileHCLBTargets(
...
+					robotIDToIPv4[s.ServerNumber] = internalIP
+					continue
+				}
+
+				klog.Warningf(
+					"%s: load balancer %s has set `use-private-ip: true`, but no InternalIP found for node %s. Continuing with ExternalIP.",
+					op,
+					svc.Name,
+					node.Name,
+				)
```
apricote, line 666: Is this event worthy?

apricote (changes requested):

internal/hcops/load_balancer.go
lukasmetzner: Just for clarification and its not a misunderstanding. Its a log message and not a k8s event. 

The log message could be helpful for debugging clusters, which are not setup correctly, but we can also discard it.

apricote: Sorry, my question is: "This is a log message right now, should this maybe also be an event?"

docs/how-to-robot-vswitch-load-balancer.md
```diff
@@ -0,0 +1,46 @@
+# How to attach load balancers to Robot private IPs
+
+With the v1.24.0 release we introduced the option to configure Internal IPs for Robot servers. This allows creating a cluster with private networks and a mixture of Robot and Cloud servers. Using the routing feature of private networks is not supported, so this requires a CNI plugin with encapsulation methods, such as Cilium with routing mode `tunnel`. Load Balancers can have targets of type IP, which can either be a public or private (vSwitch) IP of a Robot server ([API reference](https://docs.hetzner.cloud/#load-balancer-actions-add-target)).
+
+As a result, the annotation `load-balancer.hetzner.cloud/use-private-ip` can be set, if the Robot server is connected to a private network and its IP is of type [InternalIP](https://kubernetes.io/docs/reference/node/node-status/#addresses).
+
+## Configuration
+
+To configure this, enable Robot support as outlined in the [robot setup guide](./robot.md). Since the HCCM needs to fetch network data, provide the network ID using the `HCLOUD_NETWORK` environment variable.
```
apricote, line 9:
```suggestion
To configure this, enable Robot support as outlined in the [Robot setup guide](./robot.md). Since the HCCM needs to fetch network data, provide the network ID using the `HCLOUD_NETWORK` environment variable.
```

internal/hcops/load_balancer.go
lukasmetzner: Yes, I think we could add an event. I will do so.

apricote (approved):

PR #933, opened 2025-05-21
refactor: remove lb to service annotations

lukasmetzner:
The cloudprovider interface defines the service object as read-only. Therefore, we should not, and don't need to, copy data from the load balancer to the service annotations.


apricote (changes requested):
It looks like the function `AssertServiceAnnotated` in `internal/annotation/testing.go` is now unused. Its the only thing in that file, so we can just yeet it.

apricote (approved):

PR #945, opened 2025-06-06
test: use remote coverage

lukasmetzner:
We have already implemented this in the csi-driver. With this solution we can test more paths (e.g., metrics) in the code and get a more accurate test coverage.


apricote (approved):
> we can test more paths (e.g., Robot)

Do you plan to add this in a follow-up PR?

PR #957, opened 2025-06-23
483 merge the all documentation in the docs folder

lukasmetzner:


lukasmetzner:
> I'd rather revert [69a83d9](https://github.com/hetznercloud/hcloud-cloud-controller-manager/commit/69a83d955b1605bcedb394da8cbc6e018d362a9e), because it breaks the possibility for users to search the docs. Having it spread into multiple places makes it a bit difficult. No good idea on how to solve this elegantly.

Idea was to parse the `annotations/load-balancer.go` file and automatically create the Markdown table. While building this, we could link the existing pkg.go.dev docs.

The table in https://github.com/hetznercloud/hcloud-cloud-controller-manager/commit/69a83d955b1605bcedb394da8cbc6e018d362a9e was not really that great, as it contained the constant names, rather than the actual annotations you would use on a load balancer.

apricote (approved):
Very nice start 🚀

Issue #956, opened 2025-06-23
Add a value to disable the clusterrolebinding


tboerger:
### TL;DR

I'm using the helm chart for the deployment within a management cluster outside of the cluster which should be managed by the cloud-controllermanager. There should be an option to disable the clusterrolebinding and maybe even an option to prevent the hostNetwork but setting the `HCLOUD_NETWORK` and `HCLOUD_NETWORK_DISABLE_ATTACHED_CHECK`. For the latter I can workaround, the ClusterRoleBinding is a more important feature.

### Expected behavior

Disable ClusterRoleBinding for out-of-cluster usage.

lukasmetzner:
Hey,

I see your issue. I can't provide a timeframe estimate, but feel free to open a pull request.

tboerger:
Do you have a preference for the naming? Maybe `rbac.enabled` which defaults to true to stay backward compatible?

tboerger:
@lukasmetzner now I have created https://github.com/hetznercloud/hcloud-cloud-controller-manager/pull/962 to contribute this on my own.

PR #962, opened 2025-07-01
feat(helm): add value to disable ClusterRoleBinding for out-of-cluster install

tboerger:
Fixes https://github.com/hetznercloud/hcloud-cloud-controller-manager/issues/956


chart/values.yaml
```diff
@@ -129,6 +129,10 @@ robot:
   # Set to true to enable support for Robot (Dedicated) servers.
   enabled: false
 
+rbac:
+  # Enable the definition of an admin cluster role binding for the service account.
+  enabled: true
```
lukasmetzner, line 134: To align with the [official Helm recommendation](https://helm.sh/docs/chart_best_practices/rbac/?utm_source=chatgpt.com#yaml-configuration), I would suggest naming it `create`.

```suggestion
  create: true
```

tboerger: It's best practice to not using a default service account, and having the service account also has no effect if the clusterrolebinding got to be disabled for out-of-cluster deployments. But it's up to you which path we will go.

I can easily rename enabled to create. What else should I add as part of the PR?

tboerger:
Than it's up for another review @lukasmetzner :)

lukasmetzner (approved):

PR #970, opened 2025-07-09
feat: watch-based route reconciliation by default

lukasmetzner:
Enable watch-based reconciliation in the routes' controller of cloud-controller-manager, reducing API calls. Previously the routes where reconciled every ten seconds.

This feature is for an upcoming preview release, which is used to test our current upstream contribution at Kubernetes. Two forks of `cloud-provider` and `controller-manager` are used. The tag `v0.33.2` is used as a base, with the necessary patches applied manually.

This feature will be revered upon completing our tests.

Upstream WIP PR: https://github.com/kubernetes/kubernetes/pull/131220
Kubernetes Enhancement Proposal: https://github.com/kubernetes/enhancements/pull/5289
Fork of `cloud-provider`: https://github.com/lukasmetzner/cloud-provider
Fork of `controller-manager`: https://github.com/lukasmetzner/controller-manager


apricote:
> This will be blocking a way to cut patch releases, is that intended?

The code is fully gated by the feature gate, so IMO it does not necessarily block patch releases. Worst case the PR can be reverted and we can cut a patch release then.

PR #980, opened 2025-07-16
fix: revert "fix: wrong version logged on startup (#729)"

lukasmetzner:
`k8s.io/component-base/version` provides a standardized way for Kubernetes components to embed and expose their version information. Currently, our build scripts overwrite the upstream version (e.g v1.32, v1.33) with the current HCCM version (e.g v1.26, v1.27). This was done to correct a log message from `k8s.io/cloud-provider`, which reported `Version v0.0.0-master+$Format:%H$`

Our feature gate `CloudControllerManagerWatchBasedRoutesReconciliation`, which is currently tested in HCCM `v1.27.0-alpha.0`, requires Kubernetes >= v1.33. This is currently broken, as the `k8s.io` libraries will report, that the current Kubernetes version is to low.

This reverts commit 6b8cbf4f4c0e2a8e7a98bd7c8346b958e638ba57.

Using `fix` to trigger a new patch release.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -23,7 +23,6 @@ builds:
       - -s
       - -w
       - -X github.com/hetznercloud/hcloud-cloud-controller-manager/hcloud.providerVersion={{ if not .IsSnapshot }}v{{ end }}{{ .Version }}
-      - -X k8s.io/component-base/version.gitVersion={{ if not .IsSnapshot }}v{{ end }}{{ .Version }}
 
 dockers:
   - build_flag_templates: [--platform=linux/amd64]


lukasmetzner:
We might not fully revert commit 6b8cbf4f4c0e2a8e7a98bd7c8346b958e638ba57. Putting this PR in draft to keep the current PR description.

.goreleaser.yml
```diff
@@ -23,7 +23,7 @@ builds:
       - -s
       - -w
       - -X github.com/hetznercloud/hcloud-cloud-controller-manager/hcloud.providerVersion={{ if not .IsSnapshot }}v{{ end }}{{ .Version }}
-      - -X k8s.io/component-base/version.gitVersion={{ if not .IsSnapshot }}v{{ end }}{{ .Version }}
+      - -X k8s.io/component-base/version.gitVersion=v0.0.0-master+{{ if not .IsSnapshot }}v{{ end }}{{ .Version }}
```
lukasmetzner, line 26:
```suggestion
      - -X k8s.io/component-base/version.gitVersion=v0.0.0-master+v{{ .Version }}
```

@apricote Do you think we still need the `{{ if not .IsSnapshot }}v{{ end }}`?

apricote: Nah, as the new version is just inside the "build metadata" (in Semver lingo) the v prefix is not required anymore and you can remove it.

apricote (approved):
Looking good and can be merged with or without removing the "v" prefix

Issue #984, opened 2025-07-18
Support different protocols on a single loadbalancer svc


cfi2017:
### TL;DR

Currently, when an ingress controller like traefik configures a loadbalancer svc for port 80 and 443, those services are by default exposed as TCP loadbalancer services on hetzner. Unfortunately, this means the X-FORWARDED-FOR header is not correctly set.

### Expected behavior

Ideally, one could set annotations on loadbalancer services to map specific ports to specific protocols, and specific ports to specific certificates.

An additional nice-to-have feature would be the ability to sync certificates from the cluster to Hetzner, such as when cert-manager is in use.

Issue #321, opened 2022-10-24
Load Balancer per service options


WAlekseev:
Is it possible to toggle proxyprotocol: enable/disable option per service with load balancer?

If I use annotation like 

```
load-balancer.hetzner.cloud/uses-proxyprotocol: "true"
```

It forces proxy protocol on all ports described in manifest for Service type LoadBalancer.

Thanks!

morremeyer:
I'm having this exact issue right now - I'd like to be able to use proxy protocol for HTTP and HTTPS, but not for SSH on one LB.

Enabling this configuration per service (i.e. port on the Kubernetes Service object) would make this possible.

morremeyer:
Not stale. Sorry for the spam, but these auto-closing bots force this 😞

bernatvadell:
I have the same problem, I need to specify which ports use the proxy protocol.

morsik:
@morremeyer port 22 gives me GitLab vibes... if I'm correct, then check this (if you still have this issue) -> https://docs.gitlab.com/ee/administration/operations/gitlab_sshd.html?tab=Helm+chart+%28Kubernetes%29#proxy-protocol-support

This worked for me, and GitLab Shell now works correctly with PROXY protocol.

jampy:
Having the same problem, but not related to SSH.

tracure1337:
Yes please!

lukasmetzner:
The Kubernetes [ServicePort spec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.32/#serviceport-v1-core) doesn’t support custom metadata on ports, so our only option is to use an annotation to indicate which port should use the proxy protocol. I’d prefer to avoid introducing a UI that works around the limitations of the Kubernetes API.

What are your thoughts @apricote?

jtackaberry:
> I’d prefer to avoid introducing a UI that works around the limitations of the Kubernetes API.

Isn't technically the existing `load-balancer.hetzner.cloud/uses-proxyprotocol` annotation already a UI that works around the limitations of the Kubernetes API?

Annotations are, for better or worse, the canonical solution to this.  Even the [Kubernetes docs](https://kubernetes.io/docs/reference/networking/service-protocols/#protocol-proxy-special) say that you're expected to set an annotation specific to your cloud provider.

[...]

PR #985, opened 2025-07-18
feat: per-port protocol and certificates

cfi2017:
This PR implements #984 

This change introduces two annotations:
- `load-balancer.hetzner.cloud/protocol-ports` allows specifying a mapping of ports to their respective protocol
- `load-balancer.hetzner.cloud/http-certificates-ports` allows specifying a mapping of ports to one or more certificates

This PR is in a draft state as the certificates part currently returns a 422 from the Hetzner API. I have an open ticket with the loadbalancer team about this. I can replicate this behaviour through the UI.

### DIFF
--- a/docs/load_balancers.md
+++ b/docs/load_balancers.md
@@ -94,3 +94,84 @@ will delete the associated Load Balancer. If the Load Balancer is managed
 through Terraform, this causes problems. To disable this, you can enable
 deletion protection on the Load Balancer, this way hcloud-cloud-controller-manager
 will just skip deleting it when the associated `Service` is deleted.
+
+## Per-Port Protocol and Certificate Configuration
+
+The hcloud-cloud-controller-manager supports configuring different protocols and certificates for different ports of a single service using per-port annotations.
+
+### Per-Port Protocol Configuration
+
+Use the `load-balancer.hetzner.cloud/protocol-ports` annotation to specify different protocols for different ports:
+
+```yaml
+apiVersion: v1
+kind: Service
+metadata:
+  name: multi-protocol-service
+  annotations:
+    load-balancer.hetzner.cloud/protocol-ports: "80:http,443:https,9000:tcp"
+spec:
+  type: LoadBalancer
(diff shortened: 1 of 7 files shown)


cfi2017:
I'm

> Hey,
> 
> First of all, thank you very much for your contribution!
> 
> Per-port specific configuration for load balancers is currently part of a broader discussion. [This issue](https://github.com/hetznercloud/hcloud-cloud-controller-manager/issues/321#issuecomment-3095472995) summarizes the situation quite well. As mentioned there, I’ll review the options for per-port configuration and draft a proposal to ensure a consistent experience. If the team agrees with the direction, we can move forward with the implementation, and I’ll revisit your PR at that point.
> 
> The failing e2e tests on forks are expected.
> 
> Best regards, Lukas

Hey Lukas

Thanks, and I'm happy to help with implementation. For the time being I've switched to proxy protocol which fits my use-case well (forwarding IP headers would be nice, but my Ingress supports this kind of termination).

I've thought about this some more and if I were to move forward on this I'd probably rewrite this to maintain the same annotation names as upstream, but with an added port at the end of the name, eg. `load-balancer.hetzner.cloud/protocol/443: https` - you'd have multiple annotations for a mapping, but this would be easier to understand and more extendable.

cfi2017:
> > but with an added port at the end of the name, eg. load-balancer.hetzner.cloud/protocol/443: https - you'd have multiple annotations for a mapping, but this would be easier to understand and more extendable.
> 
> This will not work unfortunately, as there is only one slash allowed in the annotation to separate `key/value`. Ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set

I think my main issue with my current approach is that it's fairly convoluted. We could use a different separation character for the port, since ultimately the actual separation character does not really matter. I think that's still a cleaner solution than encoding a map in annotations, even if the Kubernetes annotation spec specifically allows for structured data in annotations. I still think it should be human-readable.

PR #1009, opened 2025-08-28
fix: remove watch based route reconciliation alpha feature

lukasmetzner:
Needs to be removed in order to support Kubernetes v1.34, as our forks are build with Kubernetes v1.33.


apricote:
Alternatively, what do you think about updating the fork instead?

lukasmetzner:
> Alternatively, what do you think about updating the fork instead?

Would be fine for me. I think we didn't properly set an end date for this test.

PR #1037, opened 2025-09-25
 feat: support Kubernetes v1.34 and drop v1.30 

lukasmetzner:
I have included dependency updates from #1008. Furthermore, the forks `cloud-provider` and `controller-manager` also got updated to v0.34.1 with our commits applied on top. This makes testing and shipping of Kubernetes v1.34 support easier.


lukasmetzner:
This needs some more work, due to the custom forks for watch based route reconciliation.

PR #1042, opened 2025-10-02
fix: add custom clusterrole for ccm helm chart

### LINKED ISSUE #1004: Stop binding to cluster-admin for clusterrole
### TL;DR

Currently the helm chart binds to the "cluster-admin", this gives the controller access to every resource in the cluster which is excessive when it only manages some services and node labels. Why not simply bind to the required resources? Is there a reason why it needs cluster-admin?

### Expected behavior

I expect the Helm chart template to be configured in such a way that it only binds to the required resources it needs to read/write/update.

hegerdes:
This adds a new `ClusterRole` to the helm ccm helm chart and implements the least privilege principle for ccm rbac permissions. 

The values are mostly taken from the [aws ccm](https://github.com/kubernetes/cloud-provider-aws/blob/master/charts/aws-cloud-controller-manager/values.yaml) but I added `ConfigMap` permissions (since hcloud ccm needs them) and removed `ServiceAccount` permissions.

The following SA permissions are not strictly needed, but can be nesessary if you want to run the ccm in an none default config where each controller gets its own SA. In these none default configs the user would need to add its own `ClusterRole`.
```yaml
  - apiGroups:
      - ""
    resources:
      - serviceaccounts
    verbs:
      - create
```

Fixes #1004

### DIFF
--- a/chart/templates/clusterrole.yml
+++ b/chart/templates/clusterrole.yml
@@ -1,0 +1,94 @@
+{{- if .Values.rbac.create }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: "system:{{ include "hcloud-cloud-controller-manager.name" . }}"
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - configmaps
+    verbs:
+      - get
+      - list
+      - watch
+  - apiGroups:
+      - ""
+    resources:
+      - events
+    verbs:
+      - create
+      - patch
+      - update
+  - apiGroups:
+      - ""
+    resources:
+      - nodes
+    verbs:
+      - "*"
+  - apiGroups:
+      - ""
+    resources:
+      - nodes/status
+    verbs:
+      - patch
+  - apiGroups:
+      - ""
+    resources:
+      - services
+    verbs:
+      - list
+      - patch
+      - update
+      - watch
+  - apiGroups:
+      - ""
+    resources:
+      - services/status
+    verbs:
+      - list
+      - patch
+      - update
+      - watch
+  - apiGroups:
+      - ""
+    resources:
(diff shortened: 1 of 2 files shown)


chart/templates/clusterrole.yml
```diff
@@ -0,0 +1,90 @@
...
+      - patch
+  - apiGroups:
+      - ""
+    resources:
+      - services
+    verbs:
+      - list
+      - patch
+      - update
+      - watch
```
lukasmetzner, line 43:
```suggestion
      - list
      - watch
```

chart/templates/clusterrole.yml
```diff
@@ -0,0 +1,90 @@
...
+  - apiGroups:
+      - ""
+    resources:
+      - services/status
+    verbs:
+      - list
+      - patch
+      - update
+      - watch
```
lukasmetzner, line 52:
```suggestion
      - patch
      - update
```

chart/templates/clusterrole.yml
```diff
@@ -0,0 +1,90 @@
...
+    resources:
+      - tokenreviews
+    verbs:
+      - create
+  - apiGroups:
+      - authorization.k8s.io
+    resources:
+      - subjectaccessreviews
+    verbs:
+      - create
```
lukasmetzner, line 89:
```suggestion
```

lukasmetzner (changes requested):
Hey @hegerdes, I have removed a few things, which are in the AWS CCM, but I don't see a reason to have them here.

Here are the officially required rules by the Kubernetes docs: https://kubernetes.io/docs/concepts/architecture/cloud-controller/#authorization-miscellaneous

In addition with the ConfigMaps, I don't see anything we would need otherwise.

PR #1052, opened 2025-10-20
feat: extend environment variables for default load balancer configuration

M4t7e:
This PR introduces additional environment variables for load balancer configuration. These variables are designed to be set globally as defaults and can be overridden using annotations.

The main motivation is to improve support for GatewayAPI, as the `Gateway` annotation limit of 8 is restrictive and many settings are commonly needed across all load balancers from the same or even differen GatewayAPI providers. Additionally, this change allows environment-specific presets such as the new [subnet IP range](https://github.com/hetznercloud/hcloud-cloud-controller-manager/pull/1031) to be set globally. This removes the need to configure these settings in each service or use templating/patching to use the same service manifest for different environments.

New environment vars:
- `HCLOUD_LOAD_BALANCERS_ALGORITHM_TYPE`
- `HCLOUD_LOAD_BALANCERS_DISABLE_PUBLIC_NETWORK`
- `HCLOUD_LOAD_BALANCERS_HEALTH_CHECK_INTERVAL`
- `HCLOUD_LOAD_BALANCERS_HEALTH_CHECK_RETRIES`
- `HCLOUD_LOAD_BALANCERS_HEALTH_CHECK_TIMEOUT`
- `HCLOUD_LOAD_BALANCERS_PRIVATE_SUBNET_IP_RANGE`
- `HCLOUD_LOAD_BALANCERS_TYPE`
- `HCLOUD_LOAD_BALANCERS_USES_PROXYPROTOCOL`

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -293,7 +293,7 @@ func (c HCCMConfiguration) Validate() (err error) {
 
 	if c.LoadBalancer.PrivateSubnetIPRange != "" {
 		if _, _, err := net.ParseCIDR(c.LoadBalancer.PrivateSubnetIPRange); err != nil {
-			errs = append(errs, fmt.Errorf("invalid value for %q: must be a valid CIDR: %v", hcloudLoadBalancersPrivateSubnetIPRange, err))
+			errs = append(errs, fmt.Errorf("invalid value for %q: must be a valid CIDR: %w", hcloudLoadBalancersPrivateSubnetIPRange, err))
 		}
 	}
 
--- a/internal/hcops/load_balancer.go
+++ b/internal/hcops/load_balancer.go
@@ -198,24 +198,26 @@ func (l *LoadBalancerOps) Create(
 	}
 
 	algType, err := annotation.LBAlgorithmType.LBAlgorithmTypeFromService(svc)
-	if err == nil {
+	switch {
+	case err == nil:
 		opts.Algorithm = &hcloud.LoadBalancerAlgorithm{Type: algType}
-	} else if errors.Is(err, annotation.ErrNotSet) {
+	case errors.Is(err, annotation.ErrNotSet):
 		if l.Cfg.LoadBalancer.AlgorithmType != "" {
 			opts.Algorithm = &hcloud.LoadBalancerAlgorithm{Type: l.Cfg.LoadBalancer.AlgorithmType}
 		}
-	} else {
+	default:
 		return nil, fmt.Errorf("%s: %w", op, err)
 	}
 
 	disablePubIface, err := annotation.LBDisablePublicNetwork.BoolFromService(svc)
-	if err == nil {
+	switch {
+	case err == nil:
 		opts.PublicInterface = hcloud.Ptr(!disablePubIface)
-	} else if errors.Is(err, annotation.ErrNotSet) {
+	case errors.Is(err, annotation.ErrNotSet):
 		if l.Cfg.LoadBalancer.DisablePublicNetwork != nil {
 			opts.PublicInterface = hcloud.Ptr(!*l.Cfg.LoadBalancer.DisablePublicNetwork)
 		}
-	} else {
+	default:
 		return nil, fmt.Errorf("%s: %w", op, err)
 	}
 
@@ -577,11 +579,12 @@ func (l *LoadBalancerOps) togglePublicInterface(ctx context.Context, lb *hcloud.
 
 	disable, err := annotation.LBDisablePublicNetwork.BoolFromService(svc)
 	var desiredDisable *bool
-	if err == nil {
+	switch {
+	case err == nil:
 		desiredDisable = hcloud.Ptr(disable)
-	} else if errors.Is(err, annotation.ErrNotSet) {
+	case errors.Is(err, annotation.ErrNotSet):
 		desiredDisable = l.Cfg.LoadBalancer.DisablePublicNetwork
-	} else {
+	default:
 		return false, fmt.Errorf("%s: %w", op, err)
 	}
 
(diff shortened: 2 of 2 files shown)


apricote:
Hey @M4t7e,

just to confirm, the "8 Annotations" limit is from the [`Gateway.spec.infrastructure.annotations`](https://gateway-api.sigs.k8s.io/reference/spec/#gatewayinfrastructure) field, right?

I found this thread where the limit was discussed and added, it sounds like this is a "soft" limit that can be raised if a reasonable case is made for more than 8 annotations: https://github.com/kubernetes-sigs/gateway-api/pull/1757#discussion_r1327843184

M4t7e:
Hey @apricote,

yes, that's the limit I was referring to. I began researching ways to preset annotations in GatewayAPI, since the concept is not to have a single Gateway (Load Balancer) for everything, like it is often the case for Ingress Controller, but to have the flexibility of creating multiple Gateways. To avoid repeating the same config, I was looking into options setting global annotation setting. That's when I came across this issue: https://github.com/kubernetes-sigs/gateway-api/issues/2734

From what I understand, the annotation limit can only be increased through provider-specific implementations like Istio and Envoy already support. In my case, I’m planning to use Cilium GatewayAPI, which as far as I know doesn’t support adding annotations with their custom config.

Btw, this is the actual issue where we want to add GatewayAPI support: https://github.com/hcloud-k8s/terraform-hcloud-kubernetes/pull/216

internal/config/config.go
```diff
@@ -183,12 +202,70 @@ func Read() (HCCMConfiguration, error) {
...
 
+	cfg.LoadBalancer.ProxyProtocolEnabled, err = getEnvBool(hcloudLoadBalancersUsesProxyProtocol, false)
+	if err != nil {
+		errs = append(errs, err)
+	}
+	// Workaround to keep bug https://github.com/hetznercloud/hcloud-cloud-controller-manager/issues/876
+	if _, exists := os.LookupEnv(hcloudLoadBalancersUsesProxyProtocol); exists {
+		cfg.LoadBalancer.ProxyProtocolEnabledSet = true
+	}
+
```
lukasmetzner: Could we make this a pointer and do a nil check instead?

internal/config/config.go
```diff
@@ -188,12 +206,66 @@ func Read() (HCCMConfiguration, error) {
...
 	disableIPv6, err := getEnvBool(hcloudLoadBalancersDisableIPv6, false)
 	if err != nil {
 		errs = append(errs, err)
 	}
 	cfg.LoadBalancer.IPv6Enabled = !disableIPv6 // Invert the logic, as the env var is prefixed with DISABLE_.
 
+	if subnetRange, ok := os.LookupEnv(hcloudLoadBalancersPrivateSubnetIPRange); ok {
+		cfg.LoadBalancer.PrivateSubnetIPRange = subnetRange
+	}
```
lukasmetzner: We could use [`net.ParseCIDR`](https://pkg.go.dev/net#ParseCIDR) in the `Validate` function to check if the format is correct.

internal/config/config.go
```diff
@@ -188,12 +206,66 @@ func Read() (HCCMConfiguration, error) {
...
+
+	if interval, ok := os.LookupEnv(hcloudLoadBalancersHealthCheckInterval); ok {
+		d, parseErr := time.ParseDuration(interval)
+		if parseErr != nil {
+			errs = append(errs, fmt.Errorf("failed to parse %s: %w", hcloudLoadBalancersHealthCheckInterval, parseErr))
+		} else {
+			cfg.LoadBalancer.HealthCheckInterval = d
+		}
+	}
```
lukasmetzner:
```suggestion
    cfg.LoadBalancer.HealthCheckInterval, err = getEnvDuration(hcloudLoadBalancersHealthCheckInterval)
    if err != nil {
		errs = append(errs, err)
    }
```

We already have a util function for this. If `errs` is not empty, HCCM won't continue and throw an error. So the `if-else` is not necessary here.

internal/config/config.go
```diff
@@ -188,12 +206,66 @@ func Read() (HCCMConfiguration, error) {
...
+
+	if timeout, ok := os.LookupEnv(hcloudLoadBalancersHealthCheckTimeout); ok {
+		d, parseErr := time.ParseDuration(timeout)
+		if parseErr != nil {
+			errs = append(errs, fmt.Errorf("failed to parse %s: %w", hcloudLoadBalancersHealthCheckTimeout, parseErr))
+		} else {
+			cfg.LoadBalancer.HealthCheckTimeout = d
+		}
+	}
```
lukasmetzner:
```suggestion
    cfg.LoadBalancer.HealthCheckTimeout, err = getEnvDuration(hcloudLoadBalancersHealthCheckTimeout)
    if err != nil {
		errs = append(errs, err)
    }
```

Same here

internal/hcops/load_balancer.go
```diff
@@ -1226,41 +1261,53 @@ func (b *hclbServiceOptsBuilder) extractHealthCheck() {
...
-		if errors.Is(err, annotation.ErrNotSet) {
+		if err == nil {
+			b.healthCheckOpts.Timeout = hcloud.Ptr(t)
+			b.addHealthCheck = true
 			return nil
-		}
-		if err != nil {
+		} else if errors.Is(err, annotation.ErrNotSet) {
+			if b.cfg.HealthCheckTimeout != 0 {
+				b.healthCheckOpts.Timeout = hcloud.Ptr(b.cfg.HealthCheckTimeout)
```
lukasmetzner:
```suggestion
			b.healthCheckOpts.Timeout = &t
			b.addHealthCheck = true
return nil
		} else if errors.Is(err, annotation.ErrNotSet) {
			if b.cfg.HealthCheckTimeout != 0 {
				b.healthCheckOpts.Timeout = &b.cfg.HealthCheckTimeout
```

M4t7e:
@lukasmetzner I made the changes you suggested. Could you please take another look?

PR #1084, opened 2025-11-17
ci: use makefile to template snapshots and deploy-manifests

lukasmetzner:
Additionally, use pre-commit to check for possible diffs in snapshots or deploy-manifests.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,7 +32,7 @@ repos:
         language: system
         entry: make -C chart lint
         pass_filenames: false
-        always_run: true
+        files: ^chart/
 
   - repo: local
     hooks:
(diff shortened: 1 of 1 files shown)


chart/Makefile
```diff
@@ -0,0 +1,54 @@
...
+default-snapshot: .snapshots/default.yaml
+	helm template hcloud-hccm . \
+		$(RELEASE_NAMESPACE) | \
+		grep -v helm.sh/chart \
+		> .snapshots/default.yaml
+
+full-snapshot: .snapshots/full.daemonset.yaml
+	helm template hcloud-hccm . \
+		$(RELEASE_NAMESPACE) \
+		-f .snapshots/full.values.daemonset.yaml \
```
jooola: Weird, I think this values file should be moved under `chart/full-values.yaml`

Then you could run a snapshot for all the values files present in the `chart/` directory (using wildcard or similar)

PR #1091, opened 2025-11-26
docs: ingress and gateway api explanation

lukasmetzner:


docs/explanation/ingress-and-gateway-api.md
```diff
@@ -0,0 +1,9 @@
+# Ingress & Gateway API
+
+Ingress and the Gateway API rely on custom resource definitions (CRDs). These CRDs require controllers that observe changes and perform the necessary actions to move the system toward the desired state.
```
apricote: Ingress does not rely on CRDs, but on the built-in `Ingress` and `IngressClass` resources (Group `networking.k8s.io/v1`)

PR #1112, opened 2025-12-17
feat: enable watch based route reconciliation by default

lukasmetzner:
This will enable the watch based route reconciliation by default. We already mentioned this feature gate in [v1.27.0](https://github.com/hetznercloud/hcloud-cloud-controller-manager/releases/tag/v1.27.0)

### DIFF
--- a/tests/e2e/cloud_test.go
+++ b/tests/e2e/cloud_test.go
@@ -172,8 +172,8 @@ func TestServiceLoadBalancersWithPrivateNetwork(t *testing.T) {
 	}
 
 	lbSvcDefinition := lbTest.ServiceDefinition(pod, map[string]string{
-		string(annotation.LBLocation):     "nbg1",
-		string(annotation.LBUsePrivateIP): "true",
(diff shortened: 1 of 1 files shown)


tests/e2e/cloud_test.go
```diff
@@ -224,88 +224,3 @@ func TestRouteNetworksPodIPsAreAccessible(t *testing.T) {
 		t.Fatal(err)
 	}
 }
-
-func TestRouteDeleteCorrectRoutes(t *testing.T) {
```
jooola: Why are we removing this?

jooola (approved):
I don't know enough to fully approve those changes, but the code change looks good.

tests/e2e/cloud_test.go
apricote: One idea for this test:

1. Manually create Routes for a made up server
2. Create `Node` object for made up node, with provider id and pod cidr set
3. HCCM will reconcile the node and delete it
4. All the events to nodes will trigger a reconcile of the routes.

Issue #1128, opened 2026-01-18
Handling of unsupported protocols can lead to `source_port_already_used`


xoxys:
### TL;DR

A service definition containing ports for unsupported protocols e.g. UDP can lead to `source_port_already_used` error if the unsupported protocol port is also used as TCP port.

### Expected behavior

Clearly warn the user about the issue (already implemented) but don't try to create the port for unsupported protocols as TCP ports as well. 

### Observed behavior

```
  ports:
  - name: http-80
    nodePort: 31257
    port: 80
    protocol: TCP
    targetPort: 10080
  - name: https-443
    nodePort: 32550
    port: 443
    protocol: TCP
    targetPort: 10443
  - name: https-443-h3
    nodePort: 32550
    port: 443
    protocol: UDP
    targetPort: 10443
```

will result in 

```
hcloud-cloud-controller-manager-5fdb6c8956-6pnbh hcloud-cloud-controller-manager I0118 08:06:37.811907       1 event.go:389] "Event occurred" object="envoy-gateway-system/envoy-public" fieldPath="" kind="Service" apiVersion="v1" type="Normal" reason="EnsuringLoadBalancer" message="Ensuring load balancer"
hcloud-cloud-controller-manager-5fdb6c8956-6pnbh hcloud-cloud-controller-manager I0118 08:06:37.929908       1 load_balancer.go:962] "update service" op="hcops/LoadBalancerOps.ReconcileHCLBServices" port=80 loadBalancerID=5549282
hcloud-cloud-controller-manager-5fdb6c8956-6pnbh hcloud-cloud-controller-manager I0118 08:06:37.994207       1 load_balancer.go:962] "update service" op="hcops/LoadBalancerOps.ReconcileHCLBServices" port=443 loadBalancerID=5549282
hcloud-cloud-controller-manager-5fdb6c8956-6pnbh hcloud-cloud-controller-manager W0118 08:06:38.063160       1 load_balancer.go:952] configured unsupported Hetzner Cloud load balancer protocol UDP for service with name envoy-public
hcloud-cloud-controller-manager-5fdb6c8956-6pnbh hcloud-cloud-controller-manager I0118 08:06:38.063226       1 load_balancer.go:973] "add service" op="hcops/LoadBalancerOps.ReconcileHCLBServices" port=443 loadBalancerID=5549282
[...]

PR #1129, opened 2026-01-18
fix: tcp service created for service with unsupported load balancer protocol

xoxys:
Fixes: https://github.com/hetznercloud/hcloud-cloud-controller-manager/issues/1128

### DIFF
--- a/internal/hcops/load_balancer.go
+++ b/internal/hcops/load_balancer.go
@@ -933,10 +933,6 @@ func (l *LoadBalancerOps) ReconcileHCLBServices(
 			err error
 		)
 
-		portNo := int(port.Port)
-		portExists := hclbListenPorts[portNo]
-		delete(hclbListenPorts, portNo)
-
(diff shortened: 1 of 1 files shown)


internal/hcops/load_balancer.go
```diff
@@ -950,6 +950,7 @@ func (l *LoadBalancerOps) ReconcileHCLBServices(
 				warnMsg,
 			)
 			klog.Warning(warnMsg)
+			continue
 		}
```
lukasmetzner:
```suggestion
portNo := int(port.Port)
portExists := hclbListenPorts[portNo]
delete(hclbListenPorts, portNo)
```

lukasmetzner (changes requested):
Hey, your change looks good. I made two suggestions. If we only put the `continue` there, the falsely created TCP services will not get cleaned up at the Hetzner Cloud Load Balancer. These changes ensure that the cleanup takes place.

Best Regards
Lukas

lukasmetzner (approved):

PR #1142, opened 2026-01-28
Add ROBOT_USE_LEGACY_PROVIDER_ID to use legacy provider ID

guettli:
For some clusters it might be challenging to
update the tool creating new nodes (like
Cluster API Provider Hetzner) and the CCM so that
both agree on the provider ID format.

If the env var ROBOT_USE_LEGACY_PROVIDER_ID is set to a true value ("1", "t", "T", "true", "TRUE",
"True"), then the CCM will use the legacy format
("hcloud://bm-NNNN") when returning
InstanceMetadata.

This applies only, when the node is created. Once
a provider ID is set, it won't be changed again.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ hcloud-cloud-controller-manager
 *.tgz
 hack/.*
 coverage/
+.vscode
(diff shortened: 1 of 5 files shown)


apricote:
Hey @guettli, what has changed since https://github.com/hetznercloud/hcloud-cloud-controller-manager/pull/1066#issuecomment-3485077616?

PR #1163, opened 2026-02-24
feat(robot): allow Robot support without API credentials for IP-based LB targets

### LINKED ISSUE #1162: feat(robot): support Robot without API credentials for IP-based LB targets
### TL;DR

When using Robot servers connected via vSwitch, the HCCM requires `ROBOT_USER` / `ROBOT_PASSWORD` at startup even though the Robot API is not needed for IP-based LB targets. The node's `InternalIP` from the Kubernetes Node object is sufficient to add targets to the Load Balancer.

Robot API credentials are account-wide and grant access to all dedicated servers (including reinstall, rescue mode). For setups that only need the service controller, this is unnecessary exposure.

### Expected behavior

`ROBOT_USER` / `ROBOT_PASSWORD` should be optional when `robot.enabled` is `true`. Without credentials, the HCCM should derive IP targets from the Kubernetes Node's `InternalIP` instead of querying the Robot API. The node controllers must be disabled by the user (`--controllers=*,-cloud-node,-cloud-node-lifecycle`).

Partial credentials (only user or only password) should be rejected as a misconfiguration.

robertlemke:
When `robot.enabled` is set but no `ROBOT_USER` / `ROBOT_PASSWORD` are provided, the HCCM now derives IP targets directly from the Kubernetes Node's `InternalIP` instead of querying the Robot API. This is useful for setups where Robot servers are connected via vSwitch and only the service controller is needed.

Existing behavior is unchanged when credentials are provided. Partial credentials (only user or only password) are rejected during validation.

Fixes: https://github.com/hetznercloud/hcloud-cloud-controller-manager/issues/1162

### DIFF
--- a/docs/explanation/robot-support.md
+++ b/docs/explanation/robot-support.md
@@ -54,4 +54,20 @@ If you absolutely need to use different names in Robot & Hostname, you can also 
 
 ## Credentials
 
+Robot API credentials (`ROBOT_USER` / `ROBOT_PASSWORD`) are **optional**. They control which features are available:
+
+### With Credentials
+
+All features described above are available: the Node Controller sets labels and addresses from the Robot API, the Node Lifecycle Controller manages shutdown detection and node deletion, and the Service Controller adds Robot servers as Load Balancer targets.
+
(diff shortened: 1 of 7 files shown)


lukasmetzner (changes requested):
Hey, I did a quick review and left some suggestions.

lukasmetzner (approved):

PR #1215, opened 2026-04-20
fix(route): error handling on locked networks

lukasmetzner:
If an error code `conflict` or `locked` is returned when adding a new route, the retry mechanism of HCCM sleeps the current goroutine for five seconds and calls the `CreateRoute` method recursively, without a max retry counter.

This behavior is inefficient and prone to error. Additionally, the `k8s.io/cloud-provider` library already performs retry handling on errors. Doing this twice might result in a fast rate-limit exhaustion.

I propose to remove retry handling from our side and let `k8s.io/cloud-provider` handle this.

### DIFF
--- a/hcloud/routes.go
+++ b/hcloud/routes.go
@@ -9,6 +9,7 @@ import (
 
 	"golang.org/x/time/rate"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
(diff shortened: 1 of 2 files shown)


lukasmetzner:
In https://github.com/hetznercloud/hcloud-cloud-controller-manager/pull/1215/changes/64547037c0a8bdf7c596dda7cb7929d802041503 I have experimented with an additional jittered retry on a `locked` or `conflict` error. For a healthy upscale path from zero to 40 nodes, I did not encounter `locked` or `conflict` errors. Feel free to leave some feedback, if think I should bring this back.

hcloud/routes.go
```diff
@@ -159,35 +157,35 @@ func (r *routes) CreateRoute(ctx context.Context, clusterName string, nameHint s
...
-		}
+	opts := hcloud.NetworkAddRouteOpts{
+		Route: hcloud.NetworkRoute{
+			Destination: cidr,
+			Gateway:     ip,
+		},
+	}
+	action, _, err := r.client.Network.AddRoute(ctx, r.network, opts)
+	if err != nil {
+		return fmt.Errorf("%s: %w", op, err)
```
apricote, line 190: The [calling code](https://github.com/kubernetes/cloud-provider/blob/1722119cb3f9a9ed49205f1dd5f95212a2576339/controllers/route/route_controller.go#L426-L432) uses `clientretry.RetryOnConflict()` to retry the call to this function if the returned error is "Conflict". Should we return this special error type here to indicate that a retry with backoff is recommended from our side?

- [`errors.NewConflict()`](https://pkg.go.dev/k8s.io/apimachinery@v0.35.4/pkg/api/errors#NewConflict)

apricote (approved):

hcloud/routes.go
lukasmetzner: I like the idea and implemented it :)

PR #1221, opened 2026-04-23
fix: routes controller on node name drift

### LINKED ISSUE #1214: Native routing: Renaming servers in in hcloud but not Kubernetes (which you can't) results in a bad time (TM)
### TL;DR

Relevant log entries from hccm:
```
E0417 10:34:19.308084       1 route_controller.go:452] Could not create route 1af3279e-37d7-4ad4-b6c1-30a2f5aaa046 10.1.131.0/24 for node worker-nbg1-1: hcloud/CreateRoute: hcops/AllServersCache.ByName: worker-nbg1-1 hcops/AllServersCache.getCache: not found
I0417 10:34:19.308205       1 event.go:389] "Event occurred" object="worker-nbg1-1" fieldPath="" kind="Node" apiVersion="v1" type="Warning" reason="FailedToCreateRoute" message="Could not create route 1af3279e-37d7-4ad4-b6c1-30a2f5aaa046 10.1.131.0/24 for node worker-nbg1-1 after 32.457156ms: hcloud/CreateRoute: hcops/AllServersCache.ByName: worker-nbg1-1 hcops/AllServersCache.getCache: not found"
```

This happened after I changed the naming scheme of our nodes in hcloud, but not in Kubernetes.  
hccm seems to use node names to figure out routes to add/keep/delete. IMO depending on a mutable property doesn't seem like a sensible approach. Would it be possible to somehow add a server-id annotation instead and use that?  
Without some kind of DaemonSet that interacts with the metadata service (like hcloud-csi) I don't see how though.

### Expected behavior

.

### Observed behavior

.

### Minimal working example

_No response_

### Log output

```plain text

```

### Additional information

_No response_

lukasmetzner:
If a Kubernetes node got successfully initialized we have the Server ID stored in the node objets `providerID`. When renaming a Server, causing a drift between the node and Server name, we should still be able to identify the Server via its ID. This is currently not implemented in the routes' controller, where we rely purly on the Servers name.

As the upstream library only provides us with the Kubernetes node name and not the full object, we have to first fetch it from the Kubernetes API. With the node object at hand we can identify and fetch the Server by its ID. Furthermore, we can use the Kubernetes node object to improve our handling of Kubernetes events. The event emitted on a CIDR mismatch is currently not associated with the affected node, because so far we did not have access to the nodes UUID.

Fixes #1214

### DIFF
--- a/hcloud/routes.go
+++ b/hcloud/routes.go
@@ -152,10 +152,10 @@ func (r *routes) resolveRouteTarget(ctx context.Context, nodeName string) (*core
 
 	id, isCloudServer, err := providerid.ToServerID(node.Spec.ProviderID)
 	if err != nil {
-		return nil, nil, fmt.Errorf("error parsing providerID %q for node %s: %w", node.Spec.ProviderID, nodeName, err)
+		return nil, nil, fmt.Errorf("error parsing provider id %q for node %s: %w", node.Spec.ProviderID, nodeName, err)
 	}
 	if !isCloudServer {
-		return nil, nil, fmt.Errorf("node %s is not a Cloud server, routes are only supported for Cloud servers", node.Name)
+		return nil, nil, fmt.Errorf("node %s is not a cloud server, routes are only supported for cloud servers", node.Name)
 	}
 
 	server, err := r.serverCache.ByID(ctx, id)
@@ -163,7 +163,7 @@ func (r *routes) resolveRouteTarget(ctx context.Context, nodeName string) (*core
 		server, err = r.serverCache.ByName(ctx, node.Name)
 	}
 	if err != nil {
-		return nil, nil, fmt.Errorf("error fetching node %s by ID: %w", nodeName, err)
+		return nil, nil, fmt.Errorf("error fetching node %s by id: %w", nodeName, err)
 	}
 
 	privNet, ok := findServerPrivateNetByID(server, r.network.ID)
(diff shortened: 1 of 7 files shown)


hcloud/routes.go
```diff
@@ -105,85 +109,132 @@ func (r *routes) CreateRoute(ctx context.Context, _ string, _ string, route *clo
...
+	metrics.OperationCalled.WithLabelValues(op).Inc()
+
+	node, err := r.nodeLister.Get(nodeName)
+	if err != nil {
+		return nil, nil, fmt.Errorf("%s: %w", op, err)
+	}
+
+	if node.Spec.ProviderID == "" {
+		return nil, nil, fmt.Errorf("%s: node %q not yet initialized", op, node.Name)
+	}
```
apricote: Is this a breaking change? IIRC it was previously possible to only use the route controller, without the node controller being active.

It feels like we talked about it sometime ago, but I forgot our resolution and there was no public doucmnetation in this PR.

lukasmetzner: I think we can just make this a soft check and try to find the server via its ID (should be our happy path) and otherwise, fallback to the old behavior.

apricote (approved):

andsens:
hot damn! Quite the short turnaround time on such a big change! I'm impressed and very grateful. Thank you :pray:  
Looking forward to updating our server naming scheme :-D

lukasmetzner:
Took the opportunity to clean up some old code :D

PR #1222, opened 2026-04-23
test: address e2e test flakiness

lukasmetzner:
- Increase Load Balancer creation timeout to prevent early test abort (previously decreased: https://github.com/hetznercloud/hcloud-cloud-controller-manager/commit/bccaddb74750e7273414fbf2ea54ae76f3ee97fe#diff-8d48bac6a439e62ea3b0740b26d2d6b9c93e1aeebc67aa9cc562abac695b13d5L291)
- If we can't cleanly destroy the namespace after three minutes we can info log the error and rely on the Servers being destroyed
- Cleanup all remaining Load Balancers when stopping the cluster

### DIFF
--- a/tests/e2e/helper_test.go
+++ b/tests/e2e/helper_test.go
@@ -14,7 +14,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/stretchr/testify/require"
 	hrobot "github.com/syself/hrobot-go"
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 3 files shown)


tests/e2e/helper_test.go
```diff
@@ -113,10 +114,20 @@ func (tc *TestCluster) Stop() error {
...
+	lbs, err := tc.hcloud.LoadBalancer.AllWithOpts(ctx, hcloud.LoadBalancerListOpts{
+		ListOpts: hcloud.ListOpts{
+			LabelSelector: selector,
+		},
+	})
+	if err != nil {
+		return fmt.Errorf("fetching load balancers via selector %s: %w", selector, err)
+	}
+	for _, lb := range lbs {
+		fmt.Printf("force-deleting leaked load balancer %d (%s)\n", lb.ID, lb.Name)
```
jooola, line 154: Should we use `t.Log`?

jooola (approved):

tests/e2e/helper_test.go
lukasmetzner: `Stop` is run after `testing.M.Run()`, so we don't have access to `testing.T` anymore. I have added a comment for clarification.

PR #1252, opened 2026-05-26
feat(instances): cache server lookups to reduce API calls

lukasmetzner:
The InstancesV2 controller looks up Servers by ID/name in every method of the controller, generating significant API traffic. This PR introduces an experimental server cache that sits between the controller and the API.


internal/servercache/allservercache.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/hetznercloud/hcloud-go/v2/hcloud"
+)
+
+var _ ServerCache = (*AllServerCache)(nil)
+
+type AllServerCache struct {
+	subsystem string
+	mode      Mode
+	ttl       time.Duration
+	expiresAt time.Time
```
jooola: I was hoping to have a single cache implementation, where the mode could be picked on the fly (or at least it would be easy to do so).

My idea was that both cache "modes" use a cache entry struct with an expiration date (as it is implement in the per server cache), and decide on the refresh actions to perform based on the mode: if mode is per server, only refresh that server, if the mode is all, refresh all servers.

This would mean with the mode "all", during a refresh iterating over all cache entries to update the `expiredAt` property.

With this in mind, we could use the same cache for both modes, and even add options in the `ByName` or `ByID` methods, to pass e.g. `cache.ByName(ctx, "my-server", WithMode("all"), WithExpiration(5 *time.Second))`

lukasmetzner: I like the idea, I will take a look and explore the implementation detail of it.

hcloud/cloud.go
```diff
@@ -50,13 +51,14 @@ const (
...
-	client      *hcloud.Client
-	robotClient hrobot.RobotClient
-	cfg         config.HCCMConfiguration
-	recorder    record.EventRecorder
-	networkID   int64
-	cidr        string
-	nodeLister  corelisters.NodeLister
+	client        *hcloud.Client
+	robotClient   hrobot.RobotClient
+	instanceCache *servercache.Cache[hcloud.Server]
```
jooola:
```suggestion
	serverCache *servercache.Cache[hcloud.Server]
```

internal/config/config.go
```diff
@@ -29,6 +30,8 @@ const (
 	robotForwardInternalIPs = "ROBOT_FORWARD_INTERNAL_IPS"
 
 	hcloudInstancesAddressFamily = "HCLOUD_INSTANCES_ADDRESS_FAMILY"
+	hcloudCacheMode              = "HCLOUD_CACHE_MODE"
+	hcloudCacheTTL               = "HCLOUD_CACHE_TTL"
```
jooola: Maybe we still want a namespace like "servers"? With "Instances" I understand the instance controller "instance_v2". While with "servers" I understand the API Servers.

We also probably have different caching options for the robot servers.

internal/config/config.go
```diff
@@ -174,6 +185,28 @@ func Read() (HCCMConfiguration, error) {
...
+	// ---- Server Cache ----
+
+	cfg.Cache = CacheConfiguration{
+		Mode: servercache.ModeAll,
+		TTL:  CacheDefaultTTL,
+	}
+
+	if mode, ok := os.LookupEnv(hcloudCacheMode); ok {
+		klog.Warningf("Experimental: %s is experimental, breaking changes may occur within minor releases.", hcloudCacheMode)
+		cfg.Cache.Mode = servercache.Mode(mode)
```
jooola: Should we validate the value provided by the user?

internal/metrics/metrics.go
```diff
@@ -37,10 +37,15 @@ var (
 		Name: "cloud_controller_manager_operations_total",
 		Help: "The total number of operation was called",
 	}, []string{"op"})
+
+	CacheRequests = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "cloud_controller_manager_cache_requests_total",
```
jooola: We probably want to say its a "server" cache.

internal/cache/cache.go
```diff
@@ -0,0 +1,300 @@
...
+}
+
+func (c *Cache[T]) getFromCache(
+	ctx context.Context,
+	subsystem string,
+	lookup func() *entry[T],
+	fetch func() (*T, error),
+	opts ...RefreshOption,
+) (*T, error) {
+	refreshOpts := newCacheRefreshOpts(c, opts...)
```
jooola, line 133: I am tempted to do the following for all functions that pass `subsystem`, just to unclutter the args:


```suggestion
func (c *Cache[T]) getFromCache(
	ctx context.Context,
	lookup func() *entry[T],
	fetch func() (*T, error),
	opts ...RefreshOption,
) (*T, error) {
	subsystem := GetSubsystem(ctx)
	refreshOpts := newCacheRefreshOpts(c, opts...)
```

And if a ctx is missing, we can/should probably add it.

internal/cache/cache.go
```diff
@@ -0,0 +1,300 @@
...
+
+	klog.V(4).InfoS("entry not found after refresh", "subsystem", subsystem)
+	return nil, nil
+}
+
+func (c *Cache[T]) refreshOne(
+	subsystem string,
+	fetch func() (*T, error),
+	ttl time.Duration,
+) error {
```
jooola, line 188: We might want to pass a ctx here:
1. Have access to the subsystem, without passing `subsystem` via the function args.
2. Not a big fan of passing a context during in the function definition, instead of from the function caller. The context hierarchy is clearer when passed via the function caller.

docs/reference/instance_cache.md
```diff
@@ -0,0 +1,26 @@
+# Instance Cache
```
jooola:
```suggestion
# Server Cache
```

internal/servercache/servercache_test.go
```diff
@@ -0,0 +1,653 @@
...
+
+func assertServer2(t *testing.T, server *hcloud.Server) {
+	t.Helper()
+	require.NotNil(t, server)
+	assert.Equal(t, int64(2), server.ID)
+	assert.Equal(t, "test2", server.Name)
+}
+
+func TestServerCacheModeAllServers(t *testing.T) {
+	apiCalls := atomic.Int64{}
```
jooola: Let's keep it simple and replace this atomic with a regular int?

jooola (commented):
First review! Looks great, only need a bit of polishing.

internal/cache/cache.go
lukasmetzner: Done :+1:

internal/cache/cache.go
lukasmetzner: I used the context for this now. As we can also drop the `subsystem` arg from `refreshAll`, this looks a bit more consistent now.

internal/metrics/metrics.go
lukasmetzner: Yep :+1:

internal/config/config.go
lukasmetzner: Yes, good idea. I added it to the validate function.

internal/config/config.go
lukasmetzner: Added a server namespace :+1:

docs/reference/server_cache.md
```diff
@@ -0,0 +1,26 @@
...
+| Name                       | Type                | Default | Description                                                                           |
+| -------------------------- | ------------------- | ------- | ------------------------------------------------------------------------------------- |
+| `HCLOUD_SERVER_CACHE_MODE` | `all \| one \| off` | `all`   | Selects the caching strategy. See [Modes](#modes) below.                              |
+| `HCLOUD_SERVER_CACHE_TTL`  | `duration`          | `10s`   | Lifetime of cached entries. Accepts any Go `time.Duration` string (e.g. `30s`, `2m`). |
+
+## Modes
+
+### `all`
+
+Fetches every Server in the project with a single `GET /servers` call and serves all subsequent `ByID` / `ByName` lookups from the resulting snapshot until the TTL expires. The snapshot is refreshed on the next lookup after expiry. On a cache miss within the TTL (e.g. a freshly created Server), one rate-limited refresh per TTL window is allowed to pick up the new Server; further misses in the same window return without an API call.
```
jooola: We may perform more GET /servers calls, e.g. with pagination. So the "single" here is not 100% accurate.

Also, I am not sure if we are actually rate limiting here.

internal/config/config.go
```diff
@@ -29,6 +30,8 @@ const (
 	robotForwardInternalIPs = "ROBOT_FORWARD_INTERNAL_IPS"
 
 	hcloudInstancesAddressFamily = "HCLOUD_INSTANCES_ADDRESS_FAMILY"
+	hcloudCacheMode              = "HCLOUD_SERVER_CACHE_MODE"
+	hcloudCacheTTL               = "HCLOUD_SERVER_CACHE_TTL"
```
jooola:
```suggestion
	hcloudServerCacheMode              = "HCLOUD_SERVER_CACHE_MODE"
	hcloudServerCacheTTL               = "HCLOUD_SERVER_CACHE_TTL"
```

docs/reference/server_cache.md
```diff
@@ -0,0 +1,26 @@
...
+> **Experimental:** Server caching is experimental, breaking changes may occur within minor releases. We believe the implementation is safe in practice — that is why it ships enabled by default (`all`). Set `HCLOUD_SERVER_CACHE_MODE=off` to opt out.
+
+The server cache reduces calls to the Hetzner Cloud API made by the `InstancesV2` controller, which looks up Servers by ID or name to reconcile Node state. The cache sits between the controller and the Hetzner Cloud API; behavior is controlled by the environment variables below.
+
+## Environment Variables
+
+| Name                       | Type                | Default | Description                                                                           |
+| -------------------------- | ------------------- | ------- | ------------------------------------------------------------------------------------- |
+| `HCLOUD_SERVER_CACHE_MODE` | `all \| one \| off` | `all`   | Selects the caching strategy. See [Modes](#modes) below.                              |
+| `HCLOUD_SERVER_CACHE_TTL`  | `duration`          | `10s`   | Lifetime of cached entries. Accepts any Go `time.Duration` string (e.g. `30s`, `2m`). |
```
jooola: Should we recommend not going above a max value? e.g. should we say, do not cache more than e.g. 1m ?

internal/config/config.go
lukasmetzner: Done :+1:

internal/metrics/metrics.go
```diff
@@ -37,10 +37,15 @@ var (
 		Name: "cloud_controller_manager_operations_total",
 		Help: "The total number of operation was called",
 	}, []string{"op"})
+
+	CacheRequests = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "cloud_controller_manager_server_cache_requests_total",
+		Help: "Total cache requests partitioned by cache name and result.",
```
jooola:
```suggestion
		Help: "Total cache requests partitioned by subsystem, mode and result.",
```

internal/config/config.go
```diff
@@ -67,10 +70,17 @@ const (
 	AddressFamilyIPv4      AddressFamily = "ipv4"
 )
 
+const CacheDefaultTTL time.Duration = 10 * time.Second
```
jooola:
```suggestion
const ServerCacheDefaultTTL time.Duration = 10 * time.Second
```

lukasmetzner: Done

jooola (approved):
♻️ 🚀

PR #1258, opened 2026-06-04
feat(chart): expose podSecurityContext and securityContext values

### LINKED ISSUE #1253: Helm chart: expose securityContext/podSecurityContext values — Deployment runs as root with a writable root filesystem
### What

The Helm chart (`chart/`, v1.31.1) renders the `hcloud-cloud-controller-manager` Deployment with no `securityContext` on either the pod or the container, and exposes no values to set one. As a result the controller runs as **root (uid 0)** with a **writable root filesystem** by default, and operators have no in-chart way to harden it.

### Why it matters

- Trips common policy/lint gates — e.g. [kube-linter](https://docs.kubelinter.io/) `run-as-non-root` and `no-read-only-root-fs`, and the Pod Security `restricted` profile.
- On hardened clusters the only workarounds today are a repo-wide lint waiver (which weakens the gate for every workload) or a post-render kustomize patch — neither of which should be necessary for a core system component.

### Request

Expose standard values so operators can set them, ideally with secure defaults, and wire them into `chart/templates/deployment.yaml` (`spec.template.spec.securityContext` + the container `securityContext`):

```yaml
podSecurityContext:
  runAsNonRoot: true
  runAsUser: 65532          # or an appropriate non-root uid
  seccompProfile:
    type: RuntimeDefault
securityContext:            # container-level
  readOnlyRootFilesystem: true
  allowPrivilegeEscalation: false
  capabilities:
    drop: ["ALL"]
```

If the CCM binary requires a writable path or a specific uid, documenting that (or providing an `emptyDir` for the writable path) would let operators enable `readOnlyRootFilesystem` safely.

### Environment

- Chart version: 1.31.1

vishwakt:
## What

Closes #1253.

Adds `podSecurityContext` (pod-level) and `securityContext` (container-level) values to the Helm chart and wires them into **both** the Deployment and DaemonSet templates (`chart/templates/deployment.yaml`, `chart/templates/daemonset.yaml`).

Secure defaults are shipped, as proposed in the issue, so the controller no longer runs as root with a writable root filesystem out of the box:

```yaml
podSecurityContext:
  runAsNonRoot: true
  runAsUser: 65532
  seccompProfile:
    type: RuntimeDefault
securityContext:            # container-level
  readOnlyRootFilesystem: true
  allowPrivilegeEscalation: false
  capabilities:
    drop: ["ALL"]
```

## Why

This satisfies the Pod Security `restricted` profile and common lint gates (kube-linter `run-as-non-root` / `no-read-only-root-fs`), removing the need for repo-wide lint waivers or post-render kustomize patches on hardened clusters.

## Notes

- **`readOnlyRootFilesystem: true` is safe**: the runtime controller does not write to its root filesystem (the only filesystem writes in the repo are in `tools/` and `internal/testsupport/`, neither of which is in the runtime path). No writable `emptyDir` is needed.
- The image runs as root (no `USER` in the Dockerfile), so `runAsNonRoot` is correctly paired with an explicit `runAsUser: 65532`.
- Both blocks are `{{- with }}`-guarded, so operators can relax or disable them by overriding the values (e.g. set to `{}`).
- This **changes behavior on upgrade** for existing installs (they flip to non-root + read-only root fs). Flagging for release-note visibility.
- Regenerated chart snapshots (`chart/.snapshots/*.yaml`) and deploy manifests (`deploy/ccm.yaml`, `deploy/ccm-networks.yaml`); `helm lint` passes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -64,7 +64,7 @@ repos:
         require_serial: false
         additional_dependencies: [prettier@3.9.4]
         files: \.(md|ya?ml)$
-        exclude: ^(CHANGELOG.md|chart/templates/.*|chart/.snapshots/.*|docs/reference/load_balancer_.*\.md|)$
+        exclude: ^(CHANGELOG.md|chart/templates/.*|chart/.snapshots/.*|deploy/.*|docs/reference/load_balancer_.*\.md|)$
 
   - repo: local
     hooks:
--- a/chart/.snapshots/default.yaml
+++ b/chart/.snapshots/default.yaml
@@ -5,6 +5,7 @@ kind: ServiceAccount
 metadata:
   name: hcloud-cloud-controller-manager
   namespace: kube-system
+
 ---
 # Source: hcloud-cloud-controller-manager/templates/clusterrole.yml
 apiVersion: rbac.authorization.k8s.io/v1
@@ -70,6 +71,7 @@ rules:
       - list
       - watch
       - update
+
 ---
 # Source: hcloud-cloud-controller-manager/templates/clusterrolebinding.yaml
 kind: ClusterRoleBinding
@@ -86,6 +88,7 @@ subjects:
   - kind: ServiceAccount
     name: hcloud-cloud-controller-manager
     namespace: kube-system
+
 ---
 # Source: hcloud-cloud-controller-manager/templates/deployment.yaml
 apiVersion: apps/v1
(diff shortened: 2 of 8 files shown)


lukasmetzner (approved):

lukasmetzner:
Looking good. Thanks for the contribution :blush:

PR #1269, opened 2026-06-16
feat(cache): replace TTL with max-age and use server cache in routes controller

lukasmetzner:
This PR replaces the TTL setting with a max age setting. Instead of storing the expired time, the cache now stores when an entry has been refreshed. Through max age an individual call to `ByID/ByName` can retrieve older entries without adding a long TTL on top of it.

The PR further introduces the server cache to the routes controller and replaces the existing cache there. This will drop all `hcops/AllServersCache/**` metrics from HCCM.

```rp-commits
feat(cache): replace TTL with max-age
feat(cache): use server cache in routes controller
```


hcloud/routes.go
```diff
@@ -156,28 +169,25 @@ func (r *routes) resolveRouteTarget(ctx context.Context, nodeName string) (*core
 		if !isCloudServer {
 			return nil, nil, fmt.Errorf("node %s is not a cloud server, routes are only supported for cloud servers", node.Name)
 		}
-		server, err = r.serverCache.ByID(ctx, id)
+		server, err = r.serverCache.ByID(ctx, id, cache.WithMaxAge(1*time.Minute))
```
jooola: I think we added this during a discussion, but I don't recall the exact reason why we added this, do you mind adding a comment on why we bumped this to 1 minute? (was it because we didn't add a rate limiter or because we didn't implement the cache invalidate feature?)

Also, if the users configure a max age > 1 minute, I don't know if we still want to honor their config (even though 1m is already quite long). Either we always override it in the routes controller (simpler) (I think this is best), or we need an opts that only increases the max age if its lower than 1m?

internal/cache/cache_test.go
```diff
@@ -65,6 +65,14 @@ func (c *testClient) CallCount() int {
...
 }
 
+func (c *testClient) RequireCallCount(expected int) {
+	c.t.Helper()
+	require.Equal(c.t, expected, c.callCount, "expected call count to be %d, got %d", expected, c.callCount)
+
+	// /!\ Reset call count until next assertion
+	c.callCount = 0
+}
+
```
jooola: This can be removed

internal/cache/cache_test.go
```diff
@@ -301,7 +309,7 @@ func TestCache_ModeOne(t *testing.T) {
 			assertCacheHasFreshServer(t, sc, srv)
 		}
 		// We evict only after ttl+hour, not after hour.
-		time.Sleep(sc.defaultTTL)
+		time.Sleep(sc.defaultMaxAge)
 		time.Sleep(time.Hour + time.Second)
```
jooola: Could we verify this again, I think this might have changed, and we actually evict at `refreshedAt + hour`

internal/cache/cache_test.go
jooola:
```suggestion
func TestCache_ModeOne_WithMaxAge(t *testing.T) {
```

internal/cache/cache_test.go
```diff
@@ -405,10 +392,10 @@ func TestCache_ModeOne_WithModeRefreshOpts(t *testing.T) {
 		require.NoError(t, err)
 		assertServer2(t, srv)
 		assert.Equal(t, "call=2", string(srv.Status))
-		assert.Equal(t, time.Now().Add(sc.defaultTTL), sc.byID[srv.ID].expiresAt)
+		assert.Equal(t, time.Now(), sc.byID[srv.ID].refreshedAt)
 
 		// Wait for expiration of Server ID 1 and 2
-		time.Sleep(sc.defaultTTL + 1)
+		time.Sleep(sc.defaultMaxAge + 1)
```
jooola: I have noticed that debugging  nanoseconds is sometimes a bit tricky, specially when trying to reason about what is happening in the tests. Could we use `+ time.Second` here?

docs/reference/server_cache.md
```diff
@@ -2,20 +2,20 @@
 
 > **Experimental:** Server caching is experimental, breaking changes may occur within minor releases. We believe the implementation is safe in practice — that is why it ships enabled by default (`all`). Set `HCLOUD_SERVER_CACHE_MODE=off` to opt out.
 
-The server cache reduces calls to the Hetzner Cloud API made by the `InstancesV2` controller, which looks up Servers by ID or name to reconcile Node state. The cache sits between the controller and the Hetzner Cloud API; behavior is controlled by the environment variables below.
+The server cache reduces calls to the Hetzner Cloud API made by the `InstancesV2` and routes controllers, which look up Servers by ID or name to reconcile Node and route state. The cache sits between the controllers and the Hetzner Cloud API; behavior is controlled by the environment variables below.
```
jooola:
```suggestion
The server cache reduces calls to the Hetzner Cloud API made by the nodes and routes controllers, which look up Servers by ID or name to reconcile Node and route state. The cache sits between the controllers and the Hetzner Cloud API; behavior is controlled by the environment variables below.
```

hcloud/routes.go
```diff
@@ -156,28 +169,25 @@ func (r *routes) resolveRouteTarget(ctx context.Context, nodeName string) (*core
...
-		server, err = r.serverCache.ByName(ctx, node.Name)
+		server, err = r.serverCache.ByName(ctx, node.Name, cache.WithMaxAge(1*time.Minute))
 		if err != nil {
 			return nil, nil, fmt.Errorf("error looking up hcloud server by name for node %s: %w", nodeName, err)
 		}
 	}
 
-	privNet, ok := findServerPrivateNetByID(server, r.network.ID)
-	if !ok {
-		r.serverCache.InvalidateCache()
```
jooola, line 172: I think we should document the reason why this was not kept. I think we agreed on returning an error, and let the reconcile loop try again later (in 10s), instead of invalidating the cache.

But I am uncertain if this mixes well with added `WithMaxAge(1*time.Minute)`.

jooola (commented):
Great! The routes controller got way simpler!

Some of the decisions were not documented, except for that, I think this is ready 🚀

hcloud/routes.go
jooola: Ah I think we said: because List Routes (which run at each reconcile) will fetch all servers with the default max age (10s), we don't need to invalidate the cache first and just let the next reconcile loop fix this.

lukasmetzner: Added a comment

internal/cache/cache_test.go
lukasmetzner: Done

internal/cache/cache_test.go
lukasmetzner: Yes, removing the first sleep and comment. We evict after `refreshedAt + hour`.

hcloud/routes.go
lukasmetzner: I have extracted it to a constant and added an explanation as I remember it 😅

hcloud/routes.go
```diff
@@ -156,28 +176,28 @@ func (r *routes) resolveRouteTarget(ctx context.Context, nodeName string) (*core
...
-			return nil, nil, fmt.Errorf("server %s (%d): network with id %d not attached to this server", server.Name, server.ID, r.network.ID)
-		}
+	// The cache returns (nil, nil) when the server does not exist (e.g. it was deleted).
+	if server == nil {
+		return nil, nil, fmt.Errorf("hcloud server for node %s not found", nodeName)
+	}
+
+	// CreateRoute may fail if the Server is not yet attached to the// CreateRoute may fail if the Server is not yet attached to the
+	// Private Network. In that case it returns an error and is retried;
+	// ListRoutes runs first and refreshes the cache.
```
jooola: Woupsy?

lukasmetzner: Whoups indeed :sweat_smile: Fix it

jooola (approved):

PR #1313, opened 2026-08-14
feat: warn on deprecated or unconfigured lb type

lukasmetzner:
This PR adds a warning log message and event, if a Load Balancer Type is deprecated. If the Load Balancer Type is unavailable an error message is returned.

This PR adds a warning log message and event, if a Load Balancer Type is not configured by a user through the annotation or a cluster-wide default via environment variables.

Load Balancer Types are cached using our generic cache built for instances.

```rp-commits
feat(load-balancer): warn on deprecated Load Balancer Type
feat(load-balancer): warn on unconfigured Load Balancer Type
```

### DIFF
--- a/hcloud/cloud.go
+++ b/hcloud/cloud.go
@@ -43,8 +43,10 @@ import (
 )
 
 const (
-	providerName     = "hcloud"
-	apiClientTimeout = 15 * time.Second
+	providerName           = "hcloud"
+	apiClientTimeout       = 15 * time.Second
+	lbTypeCacheMaxAge      = time.Hour
+	lbTypeCacheDefaultMode = cache.ModeAll
 )
 
 // providerVersion is set by the build process using -ldflags -X.
@@ -54,6 +56,7 @@ type cloud struct {
 	client      *hcloud.Client
 	robotClient hrobot.RobotClient
 	serverCache *cache.Cache[hcloud.Server]
+	lbTypeCache *cache.Cache[hcloud.LoadBalancerType]
 	cfg         config.HCCMConfiguration
 	recorder    record.EventRecorder
 	networkID   int64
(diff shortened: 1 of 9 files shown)


internal/hcops/load_balancer.go
```diff
@@ -109,6 +116,56 @@ func (l *LoadBalancerOps) GetByID(ctx context.Context, id int64) (*hcloud.LoadBa
...
+		if time.Now().After(lbType.Deprecation.Announced) {
+			utils.WarnEventLogf(
+				l.Recorder,
+				svc,
+				"LoadBalancerTypeDeprecated",
+				"load balancer type %q is deprecated and will be removed after %s.",
+				lbTypeName,
+				lbType.Deprecation.Announced.Format(time.RFC3339),
+			)
+		}
```
jooola: Maybe you are interrested in the deprecationutil package in hcloud-go? Which also handle the case when a type is not available anymore.

hcloud/cloud.go
```diff
@@ -150,11 +153,13 @@ func NewCloud(cidr string, nodeLister corelisters.NodeLister) (cloudprovider.Int
 	klog.Infof("Hetzner Cloud k8s cloud controller %s started\n", providerVersion)
 
 	serverCache := cache.NewServerCache(client, cfg.ServerCache.Mode, cfg.ServerCache.MaxAge)
+	lbTypeCache := cache.NewLoadBalancerTypeCache(client, lbTypeCacheDefaultMode, lbTypeCacheMaxAge)
```
jooola, line 156: ✨

internal/cache/lbtypecache.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/hetznercloud/hcloud-cloud-controller-manager/internal/metrics"
+	"github.com/hetznercloud/hcloud-go/v2/hcloud"
+)
+
+var lbTypeCacheRequests = prometheus.NewCounterVec(prometheus.CounterOpts{
+	Name: "cloud_controller_manager_load_balancer_type_cache_requests_total",
+	Help: "Total cache requests to the Load Balancer Types API partitioned by subsystem, mode and result.",
+}, []string{"subsystem", "mode", "result"})
```
jooola, line 18: Really not required, but if you prefer, those opts can be split into multiple parts with the different metric Opts: https://pkg.go.dev/github.com/prometheus/client_golang@v1.24.1/prometheus#Opts


```suggestion
var lbTypeCacheRequests = prometheus.NewCounterVec(prometheus.CounterOpts{
	Namespace: "cloud_controller_manager",
	Subsystem: "load_balancer_type",
	Name: "cache_requests_total",
	Help: "Total cache requests to the Load Balancer Types API partitioned by subsystem, mode and result.",
}, []string{"subsystem", "mode", "result"})
```

jooola (approved):

PR #1325, opened 2026-08-19
fix(load-balancer): apply the service UID label to existing Load Balancers

lukasmetzner:
Set the service UID after copying the existing labels so it replaces a stale
value. Unrelated labels on the Load Balancer are still preserved.

This is reachable whenever a Load Balancer is adopted with a stale UID label:
recreating a Service under the same name gives it a new UID, and
`EnsureLoadBalancer` finds the existing Load Balancer through its name fallback,
which also covers importing Load Balancers created by other means.


jooola (approved):
Nice, the fix does indeed match the behavior described in 430d0cf.

PR #1328, opened 2026-08-19
refactor: give annotations a value type

lukasmetzner:
This is needed for an upcoming refactor, where a Load Balancer spec is built
from a corev1.Service object.

Additionally, we have a more stable doc generation, as we don't have to sync
the comment text with the targeted type.


docs/reference/load_balancer_annotations.md
```diff
@@ -38,8 +38,8 @@ This page contains all annotations, which can be specified at a Service of type
 | `load-balancer.hetzner.cloud/http-sticky-sessions` | `bool` | `false` | `No` | Enables the sticky sessions feature of Hetzner Cloud HTTP Load Balancers. |
 | `load-balancer.hetzner.cloud/health-check-protocol` | `tcp \| http \| https` | `tcp` | `No` | Sets the protocol the health check should be performed over. |
 | `load-balancer.hetzner.cloud/health-check-port` | `int` | `-` | `No` | Specifies the port the health check is be performed on. |
-| `load-balancer.hetzner.cloud/health-check-interval` | `int` | `-` | `No` | Specifies the interval in which time we perform a health check in seconds. |
```
apricote, line 41: Just to be explicit: This already expected a `300s` formatted value? The code looks like this, but the docs stated `int` and `in seconds` so far.

apricote (approved):

docs/reference/load_balancer_annotations.md
lukasmetzner: Yes, this was already an issue in the docs, which is fixed now.

PR #1329, opened 2026-08-20
refactor: use a spec model for annotation extraction

lukasmetzner:
Introduce `internal/lbspec`, which resolves a Service and the cluster-wide
configuration into a Spec describing the desired state, and which builds the
API opts from it.

- An invalid annotation now fails the whole reconcile instead of only the
step that happened to read it. A typo in an HTTP annotation therefore also
stops node targets from being updated.
- The remaining 5 hclbServiceOptsBuilder op label values disappear from
`hcloud_ccm_operations_total`, and `resolveCertificates` is added.
- Certificates are resolved once per Service rather than once per port, and
on the reconcile context instead of a detached 5s timeout.
- An empty name annotation no longer tries to rename the Load Balancer to the
empty string, and an empty hostname annotation no longer publishes an empty
hostname as the ingress address.
- `private-ipv4` is parsed, so the private network address is compared as an
address rather than as the literal annotation text.

### DIFF
--- a/internal/hcops/load_balancer.go
+++ b/internal/hcops/load_balancer.go
@@ -113,15 +113,13 @@ func (l *LoadBalancerOps) GetByID(ctx context.Context, id int64) (*hcloud.LoadBa
 
 // Create creates a new Load Balancer using the Hetzner Cloud API.
 //
-// It adds annotations identifying the HC Load Balancer to svc.
-func (l *LoadBalancerOps) Create(ctx context.Context, svc *corev1.Service) (*hcloud.LoadBalancer, error) {
+// svc is only used to report events about the created Load Balancer.
+func (l *LoadBalancerOps) Create(
+	ctx context.Context, svc *corev1.Service, spec lbspec.Spec,
+) (*hcloud.LoadBalancer, error) {
 	const op = "hcops/LoadBalancerOps.Create"
 	metrics.OperationCalled.WithLabelValues(op).Inc()
 
-	spec, err := lbspec.Resolve(l.Recorder, svc, l.Cfg.LoadBalancer)
-	if err != nil {
-		return nil, fmt.Errorf("%s: %w", op, err)
-	}
 	if spec.Location == "" && spec.NetworkZone == "" {
 		return nil, fmt.Errorf("%s: neither %s nor %s set", op, annotation.LBLocation, annotation.LBNetworkZone)
 	}
@@ -162,18 +160,16 @@ func (l *LoadBalancerOps) Delete(ctx context.Context, lb *hcloud.LoadBalancer) e
 	return nil
 }
 
-// ReconcileHCLB configures the Hetzner Cloud Load Balancer to match what is
-// defined for the K8S Load Balancer svc.
-func (l *LoadBalancerOps) ReconcileHCLB(ctx context.Context, lb *hcloud.LoadBalancer, svc *corev1.Service) (bool, error) {
+// ReconcileHCLB configures the Hetzner Cloud Load Balancer to match spec.
+//
+// svc is only used to report events about the Load Balancer.
+func (l *LoadBalancerOps) ReconcileHCLB(
+	ctx context.Context, lb *hcloud.LoadBalancer, svc *corev1.Service, spec lbspec.Spec,
+) (bool, error) {
 	const op = "hcops/LoadBalancerOps.ReconcileHCLB"
 	metrics.OperationCalled.WithLabelValues(op).Inc()
 
 	var changed bool
-
-	spec, err := lbspec.Resolve(l.Recorder, svc, l.Cfg.LoadBalancer)
-	if err != nil {
-		return changed, fmt.Errorf("%s: %w", op, err)
-	}
 
 	labelSet, err := l.changeHCLBInfo(ctx, lb, spec)
 	if err != nil {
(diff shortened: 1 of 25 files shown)


lukasmetzner:
Kubernetes events example for a Service object:

```
  Normal   EnsuringLoadBalancer           2s (x3 over 17s)   service-controller               Ensuring load balancer
  Warning  InvalidLoadBalancerAnnotation  2s (x3 over 17s)   hcloud-cloud-controller-manager  Invalid Load Balancer annotation "load-balancer.hetzner.cloud/use-private-ip": strconv.ParseBool: parsing "123": invalid syntax
  Warning  InvalidLoadBalancerAnnotation  2s (x3 over 17s)   hcloud-cloud-controller-manager  Invalid Load Balancer annotation "load-balancer.hetzner.cloud/protocol": invalid protocol: httpl
  Warning  SyncLoadBalancerFailed         2s (x3 over 17s)   service-controller               Error syncing load balancer: failed to ensure load balancer: hcloud/loadBalancers.EnsureLoadBalancer: 2 Load Balancer annotation(s) are invalid
```

internal/lbspec/spec.go
```diff
@@ -0,0 +1,154 @@
...
+const DefaultType = "lb11"
+
+// LabelServiceUID is a label added to the Hetzner Cloud backend to uniquely
+// identify a load balancer managed by Hetzner Cloud Cloud Controller Manager.
+const LabelServiceUID = "hcloud-ccm/service-uid"
+
+// Spec is the desired state of the Hetzner Cloud Load Balancer for a Service.
+//
+// Pointer fields distinguish "not configured anywhere" (nil) from a configured
+// value, so we never ask the API to change something the user never set.
```
jooola, line 32: Why also use `<prop>Unset` properties too?

internal/lbspec/spec.go
```diff
@@ -0,0 +1,154 @@
...
+	CookieName     *string
+	CookieLifetime *time.Duration
+	TimeoutIdle    *time.Duration
+	RedirectHTTP   *bool
+	StickySessions *bool
+
+	// Certificates reference certificates either by ID or by name, exactly as
+	// the annotation spelled them. Resolving names to IDs needs the API and is
+	// left to the caller. Empty for a managed certificate, which the caller
+	// looks up by label.
```
apricote: "managed certificate" has a duplicate definition here. In the API there are "managed" and "uploaded" certificates. Additionally, "managed certificate" here seems to mean a "managed" certificate that HCCM created.

I suggest renaming the HCCM internal usage to something else.

Or is not possible to create "managed certificate" in the API and then reference it in the annotations?

internal/lbspec/spec.go
```diff
@@ -0,0 +1,154 @@
...
+	// Nil when unconfigured, in which case it is left alone.
+	PublicInterface *bool
+	// IPv4RDNS and IPv6RDNS are nil when unconfigured. The empty string is a
+	// valid value and resets the record.
+	IPv4RDNS *string
+	IPv6RDNS *string
+
+	// PrivateIPv4 is the address the Load Balancer should have in the private
+	// network. Nil when unconfigured.
+	PrivateIPv4 net.IP
```
jooola: What about using netip.Addr and netip.Prefix instead of the structs in the net package?

internal/hcops/load_balancer.go
```diff
@@ -118,128 +110,27 @@ func (l *LoadBalancerOps) GetByID(ctx context.Context, id int64) (*hcloud.LoadBa
...
-	}
-	opts.NetworkZone = hcloud.NetworkZone(l.Cfg.LoadBalancer.NetworkZone)
-	if v, err := annotation.LBNetworkZone.FromService(svc); err == nil {
-		opts.NetworkZone = hcloud.NetworkZone(v)
+		return nil, fmt.Errorf("%s: %w", op, err)
 	}
-	if opts.Location == nil && opts.NetworkZone == "" {
+	if spec.Location == "" && spec.NetworkZone == "" {
 		return nil, fmt.Errorf("%s: neither %s nor %s set", op, annotation.LBLocation, annotation.LBNetworkZone)
 	}
```
jooola, line 125: Shouldn't this validation be part of the lbspec package?

jooola (commented):
First pass review.

I wonder what the future of the hcops package is now, with the addition of this new lbspec package?

internal/lbspec/spec.go
lukasmetzner: We still use it in `hcloud-go`, but thats anyway a good point to abstract it away, so we can use the `netip` package in hccm. Implement in 853818af8f9997ebdd4cfb5254946b16632c91bc

internal/hcops/load_balancer.go
lukasmetzner: Currently, HCCM does not reconcile Load Balancers when annotations are removed. If we put this validation in `Resolve`, this might break existing setups, as this error will surface when updating or importing existing Load Balancers.

We could wrap it in a helper and put it in the package though.

lukasmetzner:
> I wonder what the future of the hcops package is now, with the addition of this new lbspec package?

I would love to get rid of it, as well as the `hcloud` package. I would love to have a dedicated package for each controller, so essentially one loadbalancer package.

internal/lbspec/spec.go
lukasmetzner: Unlike the other optional fields, `Name` and `Type` are required to create a Load Balancer, so a
concrete value must always exist. In contrast, if the annotation is dropped it must not update an existing Load Balancer, or we'd rename an LB back to the default and downgrade a lb31 back to lb11. Reconciling back to our default, after an annotation is dropped and making annotations like the Type required, is something we plan for the next major release.

internal/lbspec/spec.go
lukasmetzner: I renamed it to `OwnedCertificate` in 80a42434. This also changed one metric from `hcops/LoadBalancerOps.reconcileManagedCertificate` to `hcops/LoadBalancerOps.reconcileOwnedCertificate`, which needs to be communicated in the next release notes.

apricote: Nice!

apricote (approved):
Approval from my side, nice cleanup :)

PR #1343, opened 2026-09-01
feat: configure the IPv6 ExternalIP through a Node annotation

### LINKED ISSUE #1338: feat(robot): allow providing the actual IPv6 address of a Robot node
### TL;DR

HCCM derives the IPv6 `ExternalIP` of a Robot node as `<server_ipv6_net>` + `"1"`. This assumes both that Robot knows the server's IPv6 subnet, and that the host uses `::1` within it. Neither holds in general, and there is currently no way to tell HCCM which address the node actually uses.

### Expected behavior

A Robot node should be able to end up with its real IPv6 address in `.status.addresses`.

There are two situations where the derived address is wrong today:

**1. `server_ipv6_net` is empty.**

Robot models IPv6 as one /64 per server. In setups where IPv6 is not managed per server — for example when a single /64 is configured on upstream network equipment and individual hosts take addresses out of it — Robot has nothing to report, and the field is `null` for every affected server (which the Go client decodes to the empty string):

```jsonc
// GET /server/<number>
{
  "server": {
    "server_ip": "203.0.113.7",
    "server_ipv6_net": null,   // <--
    "product": "AX102",
    "dc": "HEL1-DC12",
    "ip": ["203.0.113.7"],
    "subnet": null,            // <-- no alternative source either
    "...": "..."
  }
}
```

Today HCCM appends `"1"` to that empty string and writes the literal address `"1"` to the Node. #1337 fixes the invalid value, but the result is then simply that the node has no IPv6 `ExternalIP` at all, with no way to supply the correct one.

**2. The host does not use `::1`.**

A Robot server is assigned a /64 and the operator may configure any address out of it. HCCM hardcodes the `::1` host address, so a server configured with, say, `::2` is given an `ExternalIP` it does not answer on. The Robot API does not expose which address is actually configured on the server.

The same applies to Cloud servers: `hcloudNodeAddresses` derives the host address from the `/64` the Cloud API reports, so situation 2 is not specific to Robot.

#### Possible solutions

**a) Preserve an existing IPv6 `ExternalIP` from the Node object when `server_ipv6_net` is empty.**

`robotNodeAddresses` already forwards `InternalIP`s from the Node object (`ROBOT_FORWARD_INTERNAL_IPS`, #865). Applying the same idea to the IPv6 `ExternalIP`, but only for the case where HCCM cannot derive one itself, would stay unambiguous: HCCM keeps precedence whenever Robot has data, and falls back to what is already on the Node only when it has none.

This needs no new configuration surface, covers situation 1, and composes with #1337, which makes HCCM emit nothing in exactly that case.

**b) A Node annotation that overrides the derived address**, e.g. `instance.hetzner.cloud/robot-external-ipv6`.

This additionally covers situation 2 and makes the intent explicit rather than inferred, at the cost of new API surface — HCCM currently reads annotations on Services only, never on Nodes.

**c) `--node-ip`** was considered and does not solve it: the kubelet accepts at most one address per family, and in dual-stack clusters that slot is typically already taken by the node's `InternalIP`. It also expresses an `InternalIP` rather than an `ExternalIP`. An address that HCCM does not also report causes node initialization to fail outright (`failed to get node address from cloud provider that matches ip`, see #647).

### Additional information

- Only relevant with `HCLOUD_INSTANCES_ADDRESS_FAMILY=ipv6` or `dualstack`; the default `ipv4` never enters this code path.
- Related: #1337 (invalid `"1"` address when `server_ipv6_net` is empty), #760 / #865 (`InternalIP` forwarding for Robot nodes), #647 (node-ip matching failure).
- The Robot API returns a `subnet[]` array alongside `server_ipv6_net`, which HCCM does not currently read. It is `null` on the servers affected here, so it is not a usable fallback — there is no other field in the API response that HCCM could derive the address from.

- Correction: this issue originally stated that the Cloud code path is unaffected. It is not; see the discussion in #1343.

lukasmetzner:
This adds the Node annotation `instance.hetzner.cloud/external-ipv6` to
state the address explicitly. It applies to Cloud and Robot servers, and
takes precedence over the address we derive.

We know the IPv6 subnet assigned to a server, but not the address the
server actually uses within it, so we derive the first address of the
subnet. That address is wrong for servers that use another address out of
their subnet.

Closes #1338

```rp-commits
feat(instances): configure the IPv6 ExternalIP through a Node annotation
```


apricote:
> Robot reports the IPv6 subnet assigned to a server, but not the address
> the server actually uses within it, so we derive the first address of the
> subnet. That address is wrong for servers that use another address out of
] their subnet, [...]

Can the same not happen on Cloud servers? We also use the "magically derived" `::1` there, even though the user could configure any other address from the `/64`.

https://github.com/hetznercloud/hcloud-cloud-controller-manager/blob/7e7c19c71fe7acfef358aced2bc4724045d19234/hcloud/instances.go#L251

Over the years we had many requests to let users set their own IPs and what not (for Wireguard, or other usecases). Maybe a more general flag to "disable addresses" works better, so users can just set their own addresses through kubelet flags?

Jakob3xD:
You're right — the Cloud path derives `::1` from the `/64` in the same way, so my claim in #1338 that it doesn't have this problem is wrong. I'll correct it there.

I don't think kubelet flags can replace this, though: there's no way to declare an `ExternalIP` through them, and `--node-ip` takes at most one address per family. On a dual-stack node whose IPv6 slot already holds the `InternalIP`, the public IPv6 can't be expressed at all.

The two requests also look opposite to me. "Let me set the addresses myself" (Wireguard) fits an opt-out well. This one is "keep doing everything, but one derived value is wrong" — and an opt-out replaces a wrong address with no address. Both seem worth having, but I don't think either covers the other.

Given the Cloud gap: generalize the annotation to `instance.hetzner.cloud/external-ipv6` and honour it in `hcloudNodeAddresses` as well? It hasn't shipped in a release yet, so renaming is still free.

lukasmetzner:
Oh yeah, we could switch it to an explicit `instance.hetzner.cloud/external-ipv6` for every provider type.

lukasmetzner:
Kubelet variant with `ROBOT_FORWARD_INTERNAL_IPS` might also work, but in mixed clusters this would lead to `ExternalIPs` on cloud servers and `InternalIPs` on Robot nodes. This might not be a big deal, but is definitely inconsistent.

apricote:
Using `instance.hetzner.cloud/external-ipv6` sounds good to me :)

Jakob3xD:
Sounds good. One thing that's easy to miss when moving this to the Cloud path: `hcloudNodeAddresses` takes neither the Node nor a recorder today, and `hcloudServer` doesn't carry a recorder the way `robotServer` does — so the `IgnoredExternalIPv6` event would silently not fire for Cloud nodes unless it gets threaded through `lookupServer`.

apricote (approved):

